PR #23, opened 2016-05-17
Add test case

lburgazzoli:


peter-lawrey:
I have updated this test with the expected behaviour. Can you review?

PR #27, opened 2016-07-13
Failing test for JSONWire marshallable with multiple lists.

scottkidder:
It seems this is only a bug when implementing Marshallable and writing read/writeMarshallable by hand, and does not occur when simply extending AbstractMarshallable.


peter-lawrey:
This should be fixed now.

PR #32, opened 2016-09-08
Cc appender fixes

scottkidder:
The LongReference `lastPosition` is readVolatile by every other accessor.
Unsure if the read of the nextHeader needs to be volatile or not.


RobAustin:
thanks scott

PR #41, opened 2017-01-17
Failing test for JSONWire reading text(Bytes)

scottkidder:


peter-lawrey:
The bug was In Bytes https://github.com/OpenHFT/Chronicle-Bytes/issues/34
Fixed in SNAPSHOT

peter-lawrey:
Ass always, thank you very much for the failing test. It helps enormously.

PR #46, opened 2017-07-05
Failing test for the JSONWire formatting of list field

scottkidder:
Fields of type List<Marshallable> inside a marshallable have an issue with the 'leaf' fomatting of the list items. It seems like maybe the seperator should be ',\n' but it ends up being just ','


peter-lawrey:
Can you check this update?

PR #47, opened 2017-07-27
Update BinaryWire2Test.java

MartyIX:
Underscore is not allowed as a variable in Java 9 (http://openjdk.java.net/jeps/213).


peter-lawrey:
Thank you.

PR #128, opened 2019-02-13
Polish: remove commented out debug code.

JanStureNielsen:
I'm not sure if the intent is to leave this debug code in place. If so, a refactor could likely push most of this into a test class. What do you think, @RobAustin?

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/LongValueBitSet.java
+++ b/src/main/java/net/openhft/chronicle/wire/LongValueBitSet.java
@@ -65,18 +65,6 @@ public class LongValueBitSet implements Marshallable {
     }
 
     /**
-     * Every public method must preserve these invariants.
-     */
-    // private void checkInvariants() {
-    //    assert (wordsInUse.getVolatileValue() == 0 || words[wordsInUse.getVolatileValue() - 1]
-    //             .getValue() != 0);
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/wire/LongValueBitSet.java
```diff
@@ -835,25 +748,11 @@ public void or(LongValueBitSet set) {
...
 
-        //   if (getWordsInUse() < set.getWordsInUse()) {
-        // //      ensureCapacity(set.getWordsInUse());
-        //      wordsInUse.setValue(set.getWordsInUse());
-        //  }
-
-        // Perform logical XOR on words in common
-        for (int i = 0; i < wordsInCommon; i++) {
-            final long result;
-            result = words[i].getVolatileValue() ^ set.words[i].getVolatileValue();
```
RobAustin: looking at the diff I see this code is removed, can you double check that this should be removed. All the other code in the class, I agree can be removed because its just commented out code, but this code here is different,

JanStureNielsen: This `for` loop calculates a intermediate `result` within the `for` scope, i.e., it is never used outside the scope. I don't see any side-effects from two `getVolatileValue()` calls in the calculation so I think this whole block is unused -- likely used during debugging -- and the associated tests pass. :)

PR #129, opened 2019-02-14
Polish: convert to OpenHFT @NotNull and @Nullable

JanStureNielsen:
Based on @peter-lawrey answer to this [SO question](https://stackoverflow.com/questions/54677678/2-x-notnull-too-much), this PR convert the `NotNull` and `Nullable` annotations to the those provided in OpenHFT and removes the IntelliJ annotation dependency.


RobAustin:
One reason that this is a bad idea is we often use IntelliJ to generate the @NotNull annotations. and It will end up re-adding them to methods that already have the chronicle @NotNull annotation, for this reason alone I'm going to reject this pull request. Sorry.

JanStureNielsen:
See conversation in https://github.com/OpenHFT/Chronicle-Bytes/pull/95.

PR #146, opened 2019-10-10
Support for null marshallble fields defined using abstract types 

sheinbergon:
Currently, null marshallble fields (of marshallable classes) defined using abstract types causing an exception to be thrown on high-level object unmarshalling. This incurs a peformance penalty ( exception handling/formatting, logging setup, etc) for what is considered to be a normal part of object deserialization/ unmarshalling.

This PR fixes this behavior.


pom.xml
```diff
@@ -24,7 +24,7 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <artifactId>chronicle-wire</artifactId>
-    <version>2.17.35</version>
```
RobAustin: Could you re-issue this PR without the POM.xml

Issue #176, opened 2020-07-18
LinkageError: loader (instance of  sun/misc/Launcher$AppClassLoader): attempted  duplicate class definition for name: "net/openhft/chronicle/queue/JDBCResultBinarylightMethodWriter"


JanStureNielsen:
The [Chronicle Queue JDBC test code](https://github.com/OpenHFT/Chronicle-Queue/blob/master/src/test/java/net/openhft/chronicle/queue/JDBCServiceTest.java#L47-L99):

    try (JDBCService service = new JDBCService(in, out, () -> DriverManager.getConnection("jdbc:hsqldb:file:" + file.getAbsolutePath(), "SA", ""))) {

        JDBCStatement writer = service.createWriter();
        writer.executeUpdate("CREATE TABLE tableName (\n" +
                            "name VARCHAR(64) NOT NULL,\n" +
                            "num INT\n" +
                            ")\n");

        for (int i = 1; i < (long) noUpdates; i++)
            writer.executeUpdate("INSERT INTO tableName (name, num)\n" +
                   "VALUES (?, ?)", "name", i);

            written = System.nanoTime() - start;
            AtomicLong queries = new AtomicLong();
            AtomicLong updates = new AtomicLong();
            CountingJDBCResult countingJDBCResult = new CountingJDBCResult(queries, updates);
            MethodReader methodReader = service.createReader(countingJDBCResult);
            while (updates.get() < noUpdates) {
                if (!methodReader.readOne())
                    Thread.yield();
                }
                Closeable.closeQuietly(service);
            }
        }
    }

appears to fail with:

```
java.lang.ClassNotFoundException: net.openhft.chronicle.queue.JDBCStatementBinarylightMethodWriterpackage net.openhft.chronicle.queue;
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:355)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
	at net.openhft.compiler.CachedCompiler.loadFromJava(CachedCompiler.java:163)
	at net.openhft.compiler.CachedCompiler.loadFromJava(CachedCompiler.java:76)
	at net.openhft.chronicle.wire.GenerateMethodWriter.createClass(GenerateMethodWriter.java:257)
[...]

PR #177, opened 2020-07-18
Catch a linkage error remove debug

JanStureNielsen:
I encountered #176 and Rob quickly assembled a fix. This PR is just a narrowing of Rob's fix in 85fb5bad2b6ecb999.


RobAustin:
Thanks

PR #253, opened 2021-01-12
Feature/dont assume impl class

JerryShea:
Hey @glukos please take a look - the test for this is attached to https://github.com/ChronicleEnterprise/Chronicle-FIX/issues/556 (but @Ignored). This PR fixes it. All Wire tests pass although `net.openhft.chronicle.wire.method.MethodWriterByInterfaceTest#writeReadViaImplementation` seems to be flakier on TC for some reason.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
+++ b/src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
@@ -293,13 +293,6 @@ public class GenerateMethodReader {
             Class<?> parameterType = parameterTypes[i];
 
             final String typeName = parameterType.getCanonicalName();
-
-            if (!parameterType.isPrimitive()) {
-                // this used to use ObjectUtils.implementationToUse to determine the type and mutable
-                // object but this is not appropriate, especially for an interface type - see #253
-                fields.append(format("private %s %sarg%dtype = %s.class;\n", "Class", m.getName(), i, typeName));
-            }
-
             fields.append(format("private %s %sarg%d;\n", typeName, m.getName(), i));
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
```diff
@@ -298,16 +297,10 @@ private void handleMethod(Method m, Class<?> anInterface, String instanceFieldNa
...
-                final Class<?> implToUse = ObjectUtils.implementationToUse(parameterType);
-
-                if (ReadMarshallable.class.isAssignableFrom(implToUse)) {
-                    fields.append(format("private %s %sarg%d = ObjectUtils.newInstance(%s.class);\n",
-                            typeName, m.getName(), i, implToUse.getCanonicalName()));
-                } else
-                    fields.append(format("private %s %sarg%d;\n", typeName, m.getName(), i));
+                // this used to use ObjectUtils.implementationToUse to determine the type and mutable
+                // object but this is not appropriate, especially for an interface type - see #253
+                fields.append(format("private %s %sarg%dtype = %s.class;\n", "Class", m.getName(), i, typeName));
```
glukos: The only place where %sarg%dtype field used is here:
```
            return format("%s = %s.object(checkRecycle(%s), %stype);\n", argumentName, valueInName, argumentName, argumentName);
```
As typename is now determined at the moment of code generation, we can remove the field and reduce method reader's memory footprint.

JerryShea: Yes great idea - have made that change. Please take a look!

JerryShea: Based on our emails I think this is safe to merge - I am keen to deliver to customer - thanks for feedback

PR #280, opened 2021-05-13
Downgrade log to DEBUG when over-writing an END_OF_DATA header

nicktindall:
I just left the log message in as debug when its an `END_OF_DATA` header just incase anyone is interested. Behaviour for other incomplete headers is unchanged.


JerryShea (approved):
LGTM

rogersimmons (approved):

PR #282, opened 2021-05-20
make calculation of padding amount more consistent

JerryShea:
Just keen to get another pair of eyes on this. Previously the code was inconsistent


nicktindall (approved):
:heart:

PR #306, opened 2021-09-16
Add date parsing stage using Locale.US

### LINKED ISSUE #305: Date is serialized without regard of Locale but deserialized using default locale
When writing a Date object, Wire will use [toString](https://docs.oracle.com/javase/8/docs/api/java/util/Date.html#toString--). It will use a format compatible with en_US locale.

While deserializing, Wire will mimic the same format using SimpleDateFormat, but based on the current locale, which gives deserialization errors on en_AU, not to mention ru_RU. We need to either serialize Date with SimpleDateFormat, use a constant locale for deserialization, or add a new stage when US formatter is used.

alamar:
Fixes #305

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/Wires.java
+++ b/src/main/java/net/openhft/chronicle/wire/Wires.java
@@ -651,10 +651,12 @@ public enum Wires {
 
         private static SimpleDateFormat sdf = new SimpleDateFormat("EEE MMM d HH:mm:ss zzz yyyy");
         private static SimpleDateFormat sdf2 = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.S");
+        private static SimpleDateFormat sdf3 = new SimpleDateFormat("EEE MMM d HH:mm:ss zzz yyyy", Locale.US);
 
         static {
             sdf.setTimeZone(TimeZone.getTimeZone("GMT"));
             sdf2.setTimeZone(TimeZone.getTimeZone("GMT"));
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -669,7 +671,13 @@ public static Date parseDate(ValueIn in) {
                 } catch (ParseException pe) {
                     try {
                         synchronized (sdf2) {
-                            return sdf2.parse(text);
```
JerryShea, line 672: please we need a test for this to document behaviour and the fix

src/test/java/net/openhft/chronicle/wire/WireTests.java
```diff
@@ -119,14 +116,29 @@ public void testLenientTypeLiteral() {
 
     @Test
     public void testDate() {
+        Locale defLocale = Locale.getDefault();
+
+        Locale.setDefault(Locale.ITALY);
+
         final Bytes b = Bytes.elasticByteBuffer();
```
JerryShea, line 123: should this be elasticHeap.... so we don't have to worry about releasing?

JerryShea (approved):

src/test/java/net/openhft/chronicle/wire/WireTests.java
alamar: I'm not sure but I will skip changing the whole test class code for now

PR #308, opened 2021-09-25
Fix flaky useInterceptor

LeoYimingLi:
The test ```AbstractClassGeneratorTest.useInterceptor``` fails due to a different order in string comparison. From code trace, the root cause is the method ```WireMarshaller.getAllField```,  which calls on Java's reflection API ```getDeclaredFields```, which does not return fields in a guaranteed order. 
This PR proposes to sort the returned fields and make it deterministic.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
@@ -36,6 +36,8 @@ import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.Arrays;
+import java.util.Comparator;
 
 import static net.openhft.chronicle.core.UnsafeMemory.*;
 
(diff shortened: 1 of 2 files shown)


JerryShea:
Hi @LeoYimingLi we don't see this test as flaky in our CI infrastructure and we test on a number of different OS & JVM version combinations. Does it fail for you sometimes?

LeoYimingLi:
> Hi @LeoYimingLi we don't see this test as flaky in our CI infrastructure and we test on a number of different OS & JVM version combinations. Does it fail for you sometimes?

Hi @JerryShea , I use the tool [Nondex](https://github.com/TestingResearchIllinois/NonDex) to find it flaky. The root cause is ```getDeclaredFields``` nondeterministic return in ```net.openhft.chronicle.wire.WireMarshaller.getAllField()```
The test failure can be reproduced in this way:

0.Environment needed: Maven 3.6.0 with Java 1.8.0_292

1.clone the repo:
git clone https://github.com/OpenHFT/Chronicle-Wire 
cd Chronicle-Wire

2.run with Nondex
mvn install -DskipTests
mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=net.openhft.chronicle.wire.AbstractClassGeneratorTest#useInterceptor

And the test failure records:
image

JerryShea:
Thanks @LeoYimingLi. I appreciate what you are saying, and that getDeclaredFields is not really deterministic, but we have found that the output from this method has been stable enough. In particular, this test passes on Java 17, Java 11, and on numerous versions of Java 8 (although we have not run with the exact version you detailed above). We also test with multiple JVM vendors and on multiple hardware platforms and OSes.

Changing `WireMarshaller.getAllField` would change the order that things appear when serialised with Chronicle Wire and this, while not strictly a breaking change, would surprise users of Chronicle Wire. 

I vote that we do not make this change. @RobAustin @peter-k-lawrey what do you think?

RobAustin:
@JerryShea - I agree, changing WireMarshaller.getAllField() could break more things than it fixes.

JerryShea:
Thanks @LeoYimingLi for the contribution but I'm afraid we will have to turn this down.

PR #309, opened 2021-09-26
Fix map marshallable flaky test

arianacai1997:
The old test has two orders. However, there can be more combinations than those two. I listed all the combinations and used `assertTrue` .


JerryShea:
This test is not flaky in our infrastructure. Under what circumstances does it fail for you? The test uses LinkedHashMap and TreeMap to give a predictable order

arianacai1997:
I used `nondex` and it failed in some cases. Some errors are like the following:
```
[ERROR] Failures: 
[ERROR]   MapMarshallableTest.test:31 expected:<{[one=10, two=20, three=3]0}> but was:<{[three=30, one=10, two=2]0}>
[INFO] 
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```

JerryShea:
Thanks, but I don't see how this could happen. Could you give more details
please? Which JVM was used?

arianacai1997:
Hi Jerry. The java version is `1.8.0_29`. Please see the steps I did:
`sudo apt install maven` # install maven
`mvn install -pl . -am -DskipTests` # compile the dependencies
`mvn -pl . test -Dtest=net.openhft.chronicle.wire.map.MapMarshallableTest#test` # run the tests
`mvn -pl . edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=net.openhft.chronicle.wire.map.MapMarshallableTest#test -DnondexRuns=10` # use nondex

JerryShea:
Thanks but what does nondex do to make this test fail? The test uses LinkedHashMap and TreeMap to give a predictable order - this feels like a false positive

JerryShea (changes requested):
Please demonstrate what is going wrong in the test

JerryShea (changes requested):
marking this PR as needing changes

peter-lawrey:
The expection is that the fields should appear in order of appearance, so the test should fail if this order isn't the same.

JerryShea:
I'm going to close this as won't fix as the test only fails when run in nondex

PR #315, opened 2021-10-01
Fix Flaky testDeserializeWithNestedArray

LeoYimingLi:
The test ```DefaultMarshallerTest.testDeserializeWithNestedArray``` sometimes fails due to a different order in string comparison. From code trace, the root cause is the method ```WireMarshaller.getAllField()```, which calls on Java's reflection API [getDeclaredFields()](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredFields--), which does not return fields in a guaranteed order. Thus, certain assertion will fail in some platforms or specific machines that adopt a different iteration order.

This PR proposes to compare 2 possible sequences during the test without changing the ```getDeclaredFields``` behaviour.

### DIFF
--- a/src/test/java/net/openhft/chronicle/wire/DefaultMarshallerTest.java
+++ b/src/test/java/net/openhft/chronicle/wire/DefaultMarshallerTest.java
@@ -25,6 +25,7 @@ import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 
 public class DefaultMarshallerTest extends WireTestCommon {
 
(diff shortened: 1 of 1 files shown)


LeoYimingLi:
The way to reproduce the flaky test failure:
- 0.Environment setup: Maven 3.6.0 and Java 1.8.0_292
- 1.clone the repo:
```
git clone https://github.com/OpenHFT/Chronicle-Wire 
cd Chronicle-Wire
```
- 2.run with Nondex tool
```
mvn install -DskipTests
mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=net.openhft.chronicle.wire.DefaultMarshallerTest#testDeserializeWithNestedArray
```

And the test failure screenshot:
- TWO, THREE 11

LeoYimingLi:
Oh there are some other tests detected as flaky by Nondex tool, I can also fix them in this way if you don't mind.

JerryShea:
thanks @LeoYimingLi but I personally think we are better off not changing a lot of tests for a scenario that has never happened in the real world - i.e. it only happens when the tests are run in the nondex tool.
These changes make the tests larger and harder to follow. @peter-k-lawrey @RobAustin I am keen to hear your opinions

LeoYimingLi:
Get it :D

JerryShea:
Closing as per explanation on other similar PRs

PR #310, opened 2021-09-27
Fix flaky testReserialize

LeoYimingLi:
The test ```BinaryInTextTest.testReserialize```
sometimes fails due to a different order in string comparison. From code trace, the root cause is the method ```WireMarshaller.getAllField```, which calls on Java's reflection API ```getDeclaredFields```, which does not return fields in a guaranteed order.
This PR proposes to consider 2 possible sequences in ```assertEquals``` without changing the ```getDeclaredFields``` behaviour.

PR #320, opened 2021-10-11
Flaky fix MicroLongConverter

HildoYe:
The test `MicroLongConverterTest.testMicro` some times failed due to a different order in string comparison.
I do not know the exact reason behind it, but a reasonable guess is some uncleared states caused it. I propose a solution here,
and it magically fix some other not flaky buts in the whole test suite.
The bug is raised when parsing marshallable object from a string. It can be narrow down to `WireType.java: line 84: return (T) wire.getValueIn().object();` which finally pointing to the function `public static <E> E objectMap(ValueIn in, @Nullable E using, @Nullable Class clazz, SerializationStrategy<E> strategy);` in `public enum Wires`.
After calling `using = (E) strategy.newInstanceOrNull(clazz);`, `ThreadLocal` has some leftover data. Clearing the `ThreadLocal `before calling `using` in the following few lines will greatly reduce the flakiness. I am uncertain whether it fixes the root cause, but it does not show any flakiness in the limited runs of the test. I think it also fixes a couple of test fails elsewhere.

This PR proposes to clear the state of `ThreadLocal` during object creation phase to reduce the flakiness in the code.

The way to reproduce the test is:
`mvn -pl . edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=net.openhft.chronicle.wire.MicroLongConverterTest#testMicro`

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/Wires.java
+++ b/src/main/java/net/openhft/chronicle/wire/Wires.java
@@ -469,6 +469,9 @@ public enum Wires {
         if (using == null) {
             using = (E) strategy.newInstanceOrNull(clazz);
             nullObject = using == null;
+            // Effectively clears the previous state in the ThreadLocal.
+            if (!nullObject)
+                acquireBytesForToString();
         }
         if (Throwable.class.isAssignableFrom(clazz))
             return (E) WireInternal.throwable(in, false, (Throwable) using);
--- a/src/main/java/net/openhft/chronicle/wire/Wires.java
+++ b/src/main/java/net/openhft/chronicle/wire/Wires.java
@@ -469,9 +469,6 @@ public enum Wires {
         if (using == null) {
             using = (E) strategy.newInstanceOrNull(clazz);
             nullObject = using == null;
-            // Effectively clears the previous state in the ThreadLocal.
-            if (!nullObject)
-                acquireBytesForToString();
         }
         if (Throwable.class.isAssignableFrom(clazz))
             return (E) WireInternal.throwable(in, false, (Throwable) using);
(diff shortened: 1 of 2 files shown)


JerryShea:
@peter-k-lawrey @RobAustin @nicktindall @glukos what do you think?

@HildoYe thanks for this - please leave this with us - we will need to review this carefully as this code is in use by a lot of customers

HildoYe:
> @peter-k-lawrey @RobAustin @nicktindall @glukos what do you think?
> 
> @HildoYe thanks for this - please leave this with us - we will need to review this carefully as this code is in use by a lot of customers

I'm sorry this PR doesn't solve the flakiness indeed. I tested it with different settings but the string order is still random. My previous test somehow was just lucky enough that ignored the issue completely. I'll try to figure out if this problem is solvable, and if it is similar to PR #310 . Currently the simplest way to mute the flaky warning is to acknowledge the ordering issue in the test case and adjust it accordingly.
Meanwhile this solution does affect some random seeds in the system, and in my local version some originally failed tests pass with the modification. I assume the flaky test itself should stay exactly the same and some work needs to be done to fix the flakiness instead of acknowledge the order issue in the test.

JerryShea:
@HildoYe it would be good if you could advise what nondex is doing to make this test flaky. It is not flaky for us and we test on a lot of different JVMs/platforms

HildoYe:
@JerryShea It just reruns the given test several times with different random seeds. The number can be specified with flag `-DnondexRuns=NUMBER_HERE`. The test has a failing rate around 50% though.

HildoYe:
@JerryShea 
I've upload another fix but that's only muting the flaky test warning. It simply acknowledges that the result string does not have a guaranteed sequence.
The root cause is the same as PR #130, which is `getDeclaredFields()` does not have consistent order. If you want to remove the flakiness, add a sorting function after the call to `getDeclaredFields()` might be helpful, but then some other codes might temporarily failed due to some changed order. It might also remove a bunch of flaky test as well.

JerryShea:
Based on discussions from other nondex-related PRs, I think we have to close this. We don't want to sort the fields returned from `getDeclaredFields()` as that will break existing code & change serialised form of data

PR #340, opened 2021-11-15
Only allow matching fields as BytesMarshallable

### LINKED ISSUE #334: Unreadable messages when marshalling object or generic fields containing BytesMarshallable
The following test case: https://gist.github.com/alamar/4c78d0b9bd9a2d41402662fbd6dccffa

When we have an Object/<T> field in a class, and the value of that field is BytesMarshallable but not SelfDescribingMarshallable, the type information is lost which will lead to undefined behavior upon deserialization. At least a warning should be printed in such cases.

Discovered together with @glukos

alamar:
fixes #334

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/ValueOut.java
+++ b/src/main/java/net/openhft/chronicle/wire/ValueOut.java
@@ -657,9 +657,6 @@ public interface ValueOut {
             else
                 return typedMarshallable((WriteMarshallable) value);
         }
-        if (value instanceof WriteBytesMarshallable) {
-            return bytesMarshallable((BytesMarshallable) value);
-        }
         if (value instanceof BytesStore)
             return bytes((BytesStore) value);
         if (value instanceof CharSequence)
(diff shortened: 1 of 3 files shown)


alamar:
I'm not sure this is the final solution, but it highlights the problem: we can only use BinaryMarshallable in untypedObject()

src/main/java/net/openhft/chronicle/wire/ValueOut.java
```diff
@@ -826,6 +823,9 @@ default WireOut untypedObject(@Nullable Object value) {
         }
         if (value instanceof Marshallable)
             return marshallable((Marshallable) value);
+        if (value instanceof WriteBytesMarshallable) {
```
JerryShea, line 826: minor but consider using consistent `if` pattern i.e. line above does not have opening/closing braces for single-line `if`

JerryShea (approved):

alamar:
Unfortunately, the TwoArraysTest highlights another case, such as passing an object with fields containing desired types.

Maybe we have to add a warning and not fix/disable this scenario right away.

PR #342, opened 2021-11-16
fix flaky test in DefaultMarshallerTest

CharlesZKQ:
In `net.openhft.chronicle.wire.DefaultMarshallerTest.testDeserialize` , `testDeserialize()` is a flaky test. The reason why this test is flaky is the order of the fields returned by the `getDeclatedFields()` method is not determined, see [document](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getDeclaredFields--). This method has been used internally by `toString()` function of `DMOuterClass` and `DMNestedClass`; therefore, the output string, is non-deterministic. 

I fixed it by override the `toString()` in a fixed way for both `DMOuterClass` and `DMNestedClass` so that the order of each field is determined.

### DIFF
--- a/src/test/java/net/openhft/chronicle/wire/DMNestedClass.java
+++ b/src/test/java/net/openhft/chronicle/wire/DMNestedClass.java
@@ -8,4 +8,9 @@ class DMNestedClass extends SelfDescribingMarshallable {
         this.str = str;
         this.num = num;
     }
+
+    @Override
+    public String toString() {
+        return "{ str: " + str + ", " + "num: " + num + " }\n";
+    }
 }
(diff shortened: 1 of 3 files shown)


JerryShea:
@CharlesZKQ under what circumstances did the test fail for you?

JerryShea (changes requested):
more info needed

CharlesZKQ:
> @CharlesZKQ under what circumstances did the test fail for you?

I used nondex tool to run it, with the JVM 8.

JerryShea:
Thanks @CharlesZKQ but we have been closing [nondex-related PRs](https://github.com/OpenHFT/Chronicle-Wire/pulls?q=is%3Apr+is%3Aclosed+nondex) - please see the comments made in these

PR #345, opened 2021-11-17
Allow safe round-trip of 0xFFFF (invalid character but possible char …

### LINKED ISSUE #344: Character '\0xFFFF' turned to zero after unmarshalling
If a `char` field contains `0xFFFF`, after marshalling and unmarshalling it would contain `0`.

Wire seems to treat `0xFFFF` as an unset char value, but it will overwrite the unset char field with 0 instead.

alamar:
…value)

fixes #344


alamar:
It will also allow null value to round-trip (see test) whereas it was turning into zero.

JerryShea (approved):
looks good but should we have a test for the customer report?

alamar:
@JerryShea please re-check, I have also moved other issue-related tests to a package.

JerryShea (approved):

PR #352, opened 2021-11-24
Fix failing tests, Fix #347 and Fix #346

### LINKED ISSUE #346: NullFieldMarshallingTest.testAbstractNullFieldUnmarshalledCorrectlyBinary is failing
Before #341 it didn't fail, on 5c81549982dd23b21e4c44f815d043cdd86f0b6b it will always pass.

### LINKED ISSUE #347: Fixed JSON Wire with ZoneID
see failing test - net.openhft.chronicle.wire.TestMarshallableZoneId

minborg:
The way deserialization should be done is a bit vague. This PR fixes the outstanding failing tests while trying to be as close to the old behaviour as possible.

### DIFF
--- a/src/test/java/net/openhft/chronicle/wire/ReadmeChapter1Test.java
+++ b/src/test/java/net/openhft/chronicle/wire/ReadmeChapter1Test.java
@@ -181,7 +181,7 @@ Data{message='Hello World', number=1234567890, timeUnit=NANOSECONDS, price=10.5}
         bytes.releaseLast();
         bytes2.releaseLast();
     }
-    
+
     @Test
     public void example3() {
 /*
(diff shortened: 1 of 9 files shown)


src/test/java/net/openhft/chronicle/wire/ReadmeChapter1Test.java
```diff
@@ -181,6 +182,13 @@ public void example2() {
         bytes2.releaseLast();
     }
 
+    @Test
+    public void abstr() {
```
alamar: This test should have checks maybe?

alamar (changes requested):
This change does not fix my failing test https://github.com/OpenHFT/Chronicle-Wire/pull/351
Perhaps it should

src/test/java/net/openhft/chronicle/wire/ReadmeChapter1Test.java
minborg: Ahh. This test should be removed. I will do that in a new commit

minborg:
We could fix that in a separate PR

alamar (approved):

glukos (approved):

PR #354, opened 2021-11-25
Implement safer toString method

shunfan-shao:
# Description
Test `net.openhft.chronicle.wire.TextWireTest.longConverter` will fail under [NonDex](https://github.com/TestingResearchIllinois/NonDex) which detects flakiness under non-deterministic environment. 

## To reproduce
```
mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex \
    -Dtest=net.openhft.chronicle.wire.TextWireTest#longConverter
```

## Issue
The root cause is that ```getDeclaredFields()``` is non-deterministic, which has been a concern for a while.

When calling method ```toString()``` on the ```TwoLongs```, the internal implementation [getAllFields](https://github.com/OpenHFT/Chronicle-Wire/blob/ea/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java#L138) uses the function which cause the test to fail. 
```java
public static void getAllField(@NotNull Class clazz, @NotNull Map<String, Field> map) {
    for (@NotNull Field field : clazz.getDeclaredFields()) 
    ...
```

There are several similar tests which are potentially non-deterministic. I will update them as well if this approach is acceptable.

### DIFF
--- a/src/test/java/net/openhft/chronicle/wire/TextWireTest.java
+++ b/src/test/java/net/openhft/chronicle/wire/TextWireTest.java
@@ -2103,6 +2103,12 @@ public class TextWireTest extends WireTestCommon {
         @LongConversion(HexaDecimalConverter.class)
         long hexa2;
 
+        @Override
+        public void writeMarshallable(@NotNull WireOut wire) {
+            wire.write("hexadecimal").rawText(Long.toHexString(hexadecimal));
+            wire.write("hexa2").rawText(Long.toHexString(hexa2));
+        }
+
         public TwoLongs(long hexadecimal, long hexa2) {
             this.hexadecimal = hexadecimal;
             this.hexa2 = hexa2;


JerryShea:
@shunfan-shao we will have to close this - please see the comments in [other nondex-related PRs](https://github.com/OpenHFT/Chronicle-Wire/pulls?q=is%3Apr+nondex)

shunfan-shao:
Thanks @JerryShea for the quick response. Though I do have details for the scenario where test fails besides NonDex (eg: JVM version), I just wanted to add one more word about the pull request. ```getDeclaredFields()``` has been wild known for being non-deterministic. 

A quote from the official ducoment: https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html
> Returns an array of Field objects .... The elements in the array returned are not sorted and are not in any particular order

PR #355, opened 2021-11-25
Optimize SerializeJavaLang::parseDate. Fix #353

### LINKED ISSUE #353: Optimize SerializeJavaLang::parseDate
The said method could be optimized in terms of performance.

minborg:


src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -674,34 +676,63 @@ public static WireOut writeDate(Date date, ValueOut out) {
...
                 } catch (ParseException pe) {
+                    throw new IORuntimeException(pe);
+                }
+            }
+
+            // Try the other remaining formats
+            // Todo: optimize away exception chaining
+            try {
+                // Make sure to always synchronize on these fields in exactly the order below if used elsewhere or otherwise there might be deadlocks.
+                // SDF_2 is the most likely one since it is used for serialization in writeDate()
```
nicktindall: can't we just eliminate the possibility of deadlocks if we remove the nesting?

or at least the nesting of the synchronized blocks? e.g.

```
try {
   synchronized(thing) {
      return thing.parse(text);
   }
} catch (ParseException ex) {
  // ignore
}

try {
   synchronized(thing2) {
      return thing2.parse(text);
   }
} catch (ParseException ex) {
  // ignore
}

// none parsed, throw
throw new IORuntimeException(...)


```

minborg: I wonder how I did not think about this in the first place...

src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -674,37 +676,69 @@ public static WireOut writeDate(Date date, ValueOut out) {
...
+            if (text.length() < 1) {
+                throw new IORuntimeException("At least one character (e.g. '0') must be present in order to deserialize a Date object");
+            }
+            final char firstChar = text.charAt(0);
+
+            // Check if it is a number
+            if (firstChar == '+' || firstChar == '-' || Character.isDigit(firstChar)) {
+                boolean isAllNum = true;
+                for (int i = 1; i < text.length(); i++) {
+                    if (!Character.isDigit(text.charAt(i))) {
```
alamar, line 689: This may potentially be slow with long inputs, slower than fail-fast of other formats, so I'm not sure it should not go last.

alamar (approved):

src/main/java/net/openhft/chronicle/wire/Wires.java
minborg: I think that other valid formats have spaces or `-` relatively early so this should complete quickly for valid inputs.

PR #356, opened 2021-11-26
non-matching fields with BytesMarshallable: warn, test

### LINKED ISSUE #334: Unreadable messages when marshalling object or generic fields containing BytesMarshallable

alamar:
fixes #334

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/ValueOut.java
+++ b/src/main/java/net/openhft/chronicle/wire/ValueOut.java
@@ -678,6 +678,14 @@ public interface ValueOut {
                 return typedMarshallable((WriteMarshallable) value);
         }
         if (value instanceof WriteBytesMarshallable) {
+            if (!Wires.warnedUntypedBytesOnce) {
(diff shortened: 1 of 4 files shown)


alamar:
See also #340

minborg:
Here is an example of a static initializer warner:

```
public final class SomeWarner {

    private SomeWarner() {
    }

    static {
        System.out.println("Warn...");
    }

    static void warnOnce() {
        // Warning done in static block
    }

    // Call using SomeWarner.warnOnce()
    
}

```

minborg:
The proposed solution works but consider using a static class

minborg (approved):

alamar:
@minborg 
Unfortunately, `Warner` approach will not work, since we wish to parameterize the message with the offending class name.

PR #362, opened 2021-11-29
Fix issue with typed abstract classes

minborg:


RobAustin (changes requested):
please add test ( if not already done )

RobAustin (approved):

RobAustin (approved):

Issue #373, opened 2022-01-07
Fix an issue with method reader/writers on interfaces with generics


peter-lawrey:

peter-lawrey:
replaced by #395

PR #374, opened 2022-01-07
Fix an issue with method reader/writers on interfaces with generics #373

peter-lawrey:


src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
```diff
@@ -265,8 +265,7 @@ private void handleInterface(Class<?> anInterface, String instanceFieldName, boo
             }
 
             if (!handledMethodNames.add(m.getName())) {
-                throw new IllegalStateException("MethodReader does not support overloaded methods. " +
```
JerryShea, line 268: are you changing this behaviour?

RobAustin (approved):

peter-lawrey:
This has been replaced by a better solution coming soon.

Issue #222, opened 2020-10-14
Remove chronicle.mr_overload_dont_throw system property


JerryShea:
Related to #217, I vote we remove `Jvm.getBoolean("chronicle.mr_overload_dont_throw")` property as IMHO we should always throw if we encounter an overloaded method rather than warn and do something non-deterministic.

We now handle repeated interfaces and even repeated method names in the hierarchy much better.

This will have an impact on customer code which relies on this system property. For impacted customers, the only solution is to refactor to remove overloaded methods.

peter-lawrey:
To be removed in x.22

JerryShea:
#comment If you were making use of this system property then it will now be ignored, and if you try and use MethodReaders/Writers on a class with overloaded methods, you will get an exception thrown

hft-team-city:
Released in [Chronicle-Wire-2.22ea2](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.22ea2), [BOM-2.22ea29](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22ea29)

hft-team-city:
Released in [Chronicle-Wire-2.22ea11](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.22ea11), [BOM-2.22ea69](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22ea69)

PR #375, opened 2022-01-22
De-flake/simplify JSON222Test

nicktindall:
This test occasionally failed because it used to start all the tests for each file when generating the parameters list, and sometimes if it was running under load some of these would still be running when `net.openhft.chronicle.wire.WireTestCommon#assertReferencesReleased` was called in the `@After`, an example is [here](https://teamcity.chronicle.software/buildConfiguration/OpenHFT_BuildAll_Build21/645098?showRootCauses=false&showLog=645098_67312_60799.67311.67312&expandBuildProblemsSection=true&expandBuildTestsSection=true)

Not sure if the tests NEEDED to be running in the ForkJoinPool? if so we can go back to doing that without re-introducing the flappiness.

### DIFF
--- a/src/test/java/net/openhft/chronicle/wire/issue/JSON222Test.java
+++ b/src/test/java/net/openhft/chronicle/wire/issue/JSON222Test.java
@@ -18,7 +18,6 @@
 package net.openhft.chronicle.wire.issue;
 
 import net.openhft.chronicle.bytes.Bytes;
-import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.Maths;
 import net.openhft.chronicle.core.OS;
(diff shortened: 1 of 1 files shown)


JerryShea (approved):
Does not appear to run any slower without the FJP

nicktindall:
> Does not appear to run any slower without the FJP

Yeah I wasn't worried about running speed, I was just wondering whether whatever this test is testing was something to do with being in the FJP or in parallel or something. It's in `issue` package, but [issue 222](https://github.com/OpenHFT/Chronicle-Wire/issues/222) doesn't look like it's got anything to do with this test.

There are 222 JSON files though. Would be better if it had a descriptive name.

PR #379, opened 2022-02-09
Add isPresent to ValueIn interface, 

### LINKED ISSUE #378: There's no way to tell if a value was present
Currently when reading from a `WireIn`, if you call
```
ValueIn valueIn = wireIn.read("fieldName");
```

the only way to determine if the field was present is
```
boolean wasPresent = valueIn instanceof DefaultValueIn;
```

It feels like there should be a way to do this via the `ValueIn` interface, rather than reaching under the hood in this way and relying on implementation details.

### Proposed solution
My proposed solution is to add a method `boolean isPresent()` to the `ValueIn` interface. `DefaultValueIn` could return `false` for this and every `ValueIn` that represented a present value could return `true`.

nicktindall:
also replace occurrences of `xxx instanceof DefaultValueIn`, Fixes #378

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/ValueIn.java
+++ b/src/main/java/net/openhft/chronicle/wire/ValueIn.java
@@ -548,6 +548,15 @@ public interface ValueIn {
     @Nullable
     Object objectWithInferredType(Object using, SerializationStrategy strategy, Class type);
 
+    /**
+     * Check if the value is present
+     *
+     * @return true if the value was present in the Wire, false otherwise
+     */
+    default boolean isPresent() {
+        return true;
+    }
+
     @NotNull
     default UUID uuid() {
         return UUID.fromString(text());
--- a/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
@@ -554,18 +554,16 @@ public class WireMarshaller<T> {
         protected abstract void getValue(Object o, ValueOut write, Object previous) throws IllegalAccessException;
 
         protected void readValue(Object o, Object defaults, ValueIn read, boolean overwrite) throws IllegalAccessException {
-            if (read instanceof DefaultValueIn) {
+            if (!read.isPresent()) {
                 if (overwrite && defaults != null)
                     copy(Objects.requireNonNull(defaults), o);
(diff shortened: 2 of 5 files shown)


src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -755,9 +751,10 @@ protected void getValue(Object o, @NotNull ValueOut write, Object previous) {
         @Override
         protected void setValue(Object o, @NotNull ValueIn read, boolean overwrite) {
             @NotNull StringBuilder sb = unsafeGetObject(o, offset);
-            if (sb == null)
+            if (sb == null) {
                 sb = new StringBuilder();
                 unsafePutObject(o, offset, sb);
+            }
```
nicktindall, line 757: I noticed this when I auto-formatted. The indenting seemed to match the intention, but the lack of brackets meant we always `unsafePut..` the string buffer, even if `unsafeGet..` returned non-null.

Not sure if this was intentional for some reason?

minborg: We should always use brackets in Java.

src/main/java/net/openhft/chronicle/wire/ValueIn.java
```diff
@@ -548,6 +548,15 @@ default Object typePrefixOrObject(Class tClass) {
     @Nullable
     Object objectWithInferredType(Object using, SerializationStrategy strategy, Class type);
 
+    /**
+     * Is the value present?
```
minborg: Returns if a value is present.

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -554,18 +554,16 @@ protected void copy(Object from, Object to) throws IllegalAccessException {
         protected abstract void getValue(Object o, ValueOut write, Object previous) throws IllegalAccessException;
 
         protected void readValue(Object o, Object defaults, ValueIn read, boolean overwrite) throws IllegalAccessException {
-            if (read instanceof DefaultValueIn) {
+            if (!read.isPresent()) {
```
minborg, line 557: This is much better!

minborg (approved):

Issue #380, opened 2022-02-09
Wire to use the ClassLookup it holds if set, instead of the global ClassAliasPool


peter-lawrey:

hft-team-city:
Released in [Chronicle-Wire-2.22ea15](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.22ea15), [BOM-2.22ea121](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.22ea121)

PR #381, opened 2022-02-09
Wire to use the ClassLookup it holds if set, instead of the global ClassAliasPool #380

peter-lawrey:


pom.xml
```diff
@@ -185,7 +185,7 @@
                         <configuration>
                             <referenceVersion>2.22ea1</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>99</binaryCompatibilityPercentageRequired>
+                            <binaryCompatibilityPercentageRequired>94</binaryCompatibilityPercentageRequired>
```
RobAustin, line 188: this is not cool :-(

RobAustin (approved):

pom.xml
minborg: I think we should document every wilful API break in the POM using comments when we reduce the percentage required. Some API changes (like making an abstract class' constructor protected rather than public) are safe and others are not.

alamar: We should just put 2.23ea0 there, after all, it's 2.23 time?

PR #382, opened 2022-02-09
Fix breaks introduced by ClassNotFoundRuntimeException

nicktindall:
This commit https://github.com/OpenHFT/Chronicle-Core/commit/805d872c64d512e9da12e680d5b3507b53a9071b changed the method signature to some methods used by Wire, this fixes the try/catch blocks to compile again with the change.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/JSONWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/JSONWire.java
@@ -400,12 +400,8 @@ public class JSONWire extends TextWire {
                 final StringBuilder sb = Wires.acquireStringBuilder();
                 sb.setLength(0);
                 this.wireIn().read(sb);
-                try {
-                    final Class<?> clazz = classLookup().forName(sb.subSequence(1, sb.length()));
-                    return parseType(null, clazz);
-                } catch (ReflectiveOperationException e) {
-                    throw new RuntimeException(e);
-                }
+                final Class<?> clazz = classLookup().forName(sb.subSequence(1, sb.length()));
+                return parseType(null, clazz);
             }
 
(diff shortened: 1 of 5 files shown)


src/main/java/net/openhft/chronicle/wire/JSONWire.java
```diff
@@ -400,12 +400,8 @@ private Object parseType() {
...
                 sb.setLength(0);
                 this.wireIn().read(sb);
-                try {
-                    final Class<?> clazz = classLookup().forName(sb.subSequence(1, sb.length()));
-                    return parseType(null, clazz);
-                } catch (ReflectiveOperationException e) {
-                    throw new RuntimeException(e);
-                }
+                final Class<?> clazz = classLookup().forName(sb.subSequence(1, sb.length()));
+                return parseType(null, clazz);
```
nicktindall, line 404: No need to rethrow as RuntimeException now, it will just throw a more specific RuntimeException

src/main/java/net/openhft/chronicle/wire/JSONWire.java
```diff
@@ -460,24 +456,20 @@ private Object parseType() {
...
+                consumePadding();
+                final char endBracket = bytes.readChar();
+                assert endBracket == '}' : "Missing end bracket }, got " + endBracket + " from " + bytes;
+                consumePadding(1);
 
-                    return result;
-                } catch (ReflectiveOperationException e) {
-                    throw new RuntimeException(e);
-                }
+                return result;
```
nicktindall, line 472: Again no need to rethrow here

JerryShea (approved):

Issue #390, opened 2022-02-23
Improved comments in HexDumpBytes


peter-lawrey:
Fix some minor issues made obvious by improved recording

PR #391, opened 2022-02-23
Improved comments in HexDumpBytes #390

peter-lawrey:
Also I updated the README examples

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/ValueOut.java
+++ b/src/main/java/net/openhft/chronicle/wire/ValueOut.java
@@ -967,6 +967,12 @@ public interface ValueOut {
         return text(x);
     }
 
+    default WireOut writeInt(IntConverter intConverter, int i) {
+        StringBuilder sb = Wires.acquireStringBuilder();
+        intConverter.append(sb, i);
+        return rawText(sb);
+    }
+
     default WireOut writeLong(LongConverter longConverter, long l) {
         StringBuilder sb = Wires.acquireStringBuilder();
         longConverter.append(sb, l);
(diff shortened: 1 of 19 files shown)


src/test/java/net/openhft/chronicle/wire/issue/JSON222Test.java
```diff
@@ -119,6 +119,8 @@ public void testJSON() throws IOException {
                 if (expected.contains("\r\n"))
                     expected = expected.replaceAll("\r\n", "\n");
                 String actual = bytes2.toString();
+                // TODO FIX
```
JerryShea: should we have an issue for this?

peter-lawrey: #392

src/main/java/net/openhft/chronicle/wire/ValueOut.java
```diff
@@ -967,6 +967,12 @@ default WireOut writeString(CharSequence x) {
         return text(x);
     }
 
+    default WireOut writeInt(IntConverter intConverter, int i) {
+        StringBuilder sb = Wires.acquireStringBuilder();
```
RobAustin, line 971: maybe use a different -> Wires.acquireStringBuilder() instance

RobAustin (approved):

src/main/java/net/openhft/chronicle/wire/ValueOut.java
peter-lawrey: Wires.acquireStringBuilder() is no longer public however Wires.acquireBytes() is #393

PR #394, opened 2022-02-24
Use an internal Bytes instead of a public one https://github.com/Open…

peter-lawrey:
…HFT/Chronicle-Wire/issues/393

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireInternal.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireInternal.java
@@ -24,6 +24,7 @@ import net.openhft.chronicle.core.io.IORuntimeException;
 import net.openhft.chronicle.core.pool.EnumInterner;
 import net.openhft.chronicle.core.pool.StringBuilderPool;
 import net.openhft.chronicle.core.pool.StringInterner;
+import net.openhft.chronicle.core.threads.ThreadLocalHelper;
 import net.openhft.chronicle.core.util.*;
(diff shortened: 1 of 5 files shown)


src/main/java/net/openhft/chronicle/wire/WireInternal.java
```diff
@@ -48,7 +49,7 @@
     static final ThreadLocal<WeakReference<Bytes>> BYTES_TL = new ThreadLocal<>();
     static final ThreadLocal<WeakReference<Bytes>> BYTES_F2S_TL = new ThreadLocal<>();
     static final ThreadLocal<WeakReference<Wire>> BINARY_WIRE_TL = new ThreadLocal<>();
-    static final ThreadLocal<WeakReference<Bytes>> ABYTES_TL = new ThreadLocal();
+    static final ThreadLocal<WeakReference<Bytes>> INTERNAL_BYTES_TL = new ThreadLocal<>();
```
RobAustin, line 52: how is this closed, resource leaks...?

RobAustin (approved):

src/main/java/net/openhft/chronicle/wire/WireInternal.java
peter-lawrey: They are closed when the thread dies if it's a CleaningThread, or on a GC.

minborg: I wonder how ThreadLocals scales with Loom looming around the corner...

PR #401, opened 2022-02-25
 Added support in MethodReader for meta data AND data handlers

peter-lawrey:
@JerryShea In particular I dropped the hard-wired assumption that header/index/index2index/roll events are ignored as this is queue specific, and I need an implementation that doesn't ignore the header.


peter-lawrey:
Note: Build / Build (pull_request)  is picking up old code, Snapshot Linux x64 (Chronicle Wire) passes fine

Issue #402, opened 2022-03-02
Robust YAML parsing


alamar:
YAML parser should be more compliant with other YAML parsers, taking [https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) as an example.

hft-team-city:
Released in [Chronicle-Wire-2.23ea3](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.23ea3), [BOM-2.23ea16](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea16)

hft-team-city:
Released in [Chronicle-Wire-2.23ea3](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.23ea3), [BOM-2.23ea17](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea17)

PR #403, opened 2022-03-02
Improvements of YAML parser based on Kubernetes config #402

alamar:


alamar:
@peter-lawrey @minborg @JerryShea please review this change - I realize there's a lot of stuff going on, but actual code change is not that great. If you have concerns about any changes, I can move them to a separate PR.

pom.xml
```diff
@@ -185,7 +185,7 @@
                         <configuration>
                             <referenceVersion>2.23ea0</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>99.9</binaryCompatibilityPercentageRequired>
+                            <binaryCompatibilityPercentageRequired>99.8</binaryCompatibilityPercentageRequired>
```
minborg, line 188: Good to list the API changes in a comment here every time the % is reduced

src/main/java/net/openhft/chronicle/wire/TextWire.java
```diff
@@ -1177,6 +1190,7 @@ public TextWire trimFirstCurly(boolean trimFirstCurly) {
 
     enum NoObject {NO_OBJECT}
 
+    @Deprecated(/* To be removed and replaced by YamlWire.TextValueOut in 2.24 #411 */)
```
minborg, line 1197: Add JavaDoc deprecated

src/main/java/net/openhft/chronicle/wire/WireType.java
```diff
@@ -483,7 +481,7 @@ else if (marshallable instanceof Serializable)
      */
     @Nullable
     public <T> T fromString(@NotNull CharSequence cs) {
-        return (T) fromString(Object.class, cs);
+        return (T) fromString(null, cs);
```
minborg: Looks suspicious with `null`?

minborg (approved):
Great with many tests added.

src/main/java/net/openhft/chronicle/wire/TextWire.java
alamar: Do we actually do this? I have checked our code base and it does not look like. There is no relevant information for end-user since we are going to replace it with the same implementation, just unduplicated.

pom.xml
alamar: I'll do that in a commit message.

src/main/java/net/openhft/chronicle/wire/WireType.java
alamar: Will add comment

Issue #405, opened 2022-03-08
Improve efficiency of (de)marshalling byte[]


peter-lawrey:

hft-team-city:
Released in [Chronicle-Wire-2.23ea1](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.23ea1), [BOM-2.23ea5](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea5)

hft-team-city:
Released in [Chronicle-Wire-2.23ea2](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.23ea2), [BOM-2.23ea6](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea6)

PR #406, opened 2022-03-08
Improve efficiency of (de)marshalling byte[] #405

peter-lawrey:


src/main/java/net/openhft/chronicle/wire/BinaryWire.java
```diff
@@ -2699,7 +2699,7 @@ public WireIn bytes(@NotNull ReadBytesMarshallable bytesConsumer) {
 
         @NotNull
         @Override
-        public byte @NotNull [] bytes() {
+        public byte @NotNull [] bytes(byte[] using) {
```
RobAustin, line 2702: found `byte[] using` be @NotNull as well

RobAustin: is it a good idea to make the method have a `using` that takes an array[]. as they are a fixed size ( maybe smaller than you require ), maybe the method should not be public but packaged local.

RobAustin (approved):

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
peter-lawrey: This `using` only works if the size is the same. 
In some benchmarks at least, this is the case 100% of the time.

peter-lawrey: The better solution is to use `Bytes`

PR #417, opened 2022-03-18
Avoid producing garbage in escaping quotes

### LINKED ISSUE #377: TextWire.getEscapingQuotes() generates garbage
In my application I use Chronicle Wire and parse JSON messages containing many string-escaped numbers (i.e. `"0.1234"`). It is the same use case as shown in example under https://chronicle.software/how-to-build-low-latency-crypto-trading-system-using-java-and-chronicle-services/. 

```
    // Binance provides prices / qtys in escape quotes, so we have to read a text first
    private double readDoubleInQuotes(ValueIn valueIn) {
        doubleText.setLength(0);
        valueIn.text(doubleText);
        // StringUtils from Chronicle-Services allows to avoid allocating a String
        return StringUtils.parseDouble(doubleText);
    }
```

Profiling my application I have identified, that the `TextWire.getEscapingQuotes()` method allocates new memory for `StopCharTesters.QUOTES::escaping` supplier. I have tested with Java 11 and Java 17 with similar results. Here is the snapshot from the Async Profiler using Java 17:

CleanShot 2022-01-31 at 09 03 36@2x

alamar:
fixes #377

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/CSVWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/CSVWire.java
@@ -27,6 +27,7 @@ import java.io.Serializable;
 import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Supplier;
 
 /**
  * YAML Based wire format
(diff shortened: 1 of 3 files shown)


glukos (approved):

glukos:
Another question is why we need to store stop char testers by WeakReference, it also can cause allocations. But of a smaller magintude - in case the code is hot.

alamar:
@glukos I took care of it by filing OpenHFT/Chronicle-Core#361

Issue #414, opened 2022-03-15
Consider warning if a final variable is changed via readMarshallable


rogersimmons:
readMarshallable allows final variables to be changed - eg when reading config on startup.
As Java allows aggressive optimisation of final variables this can result in a situation where a variable in an object has one value (changed from config), but accessors will return a different value (a default baked in at compilation).

Ideally a warning should be emitted when a final variable is changed, or the config refused altogether.

RobAustin:
We should be more abrupt and throw an exception, the problem with WARNings is they get missed!

alamar:
Let's do warning now, error in 2.24?

alamar:
Turns out we have more than 10 tests which do exactly that - overwrite final fields.

What should we do - should we update them all?

@rogersimmons @RobAustin

RobAustin:
@alamar this looks done, if this is not the case please reopen

alamar:
Nope, no real work is done yet, reopening.

PR #420, opened 2022-03-23
Issue a warning when final field is overwritten #414

alamar:
#414

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
@@ -450,6 +450,7 @@ public class WireMarshaller<T> {
 
         Comment commentAnnotation;
         Boolean isLeaf;
+        boolean isFinalNoWarning;
 
         FieldAccess(@NotNull Field field) {
             this(field, null);
@@ -461,6 +462,10 @@ public class WireMarshaller<T> {
             offset = unsafeObjectFieldOffset(field);
             key = field::getName;
             this.isLeaf = isLeaf;
+
+            if ((field.getModifiers() & Modifier.FINAL) != 0)
+                isFinalNoWarning = true;
+
             try {
                 commentAnnotation = field.getAnnotation(Comment.class);
             } catch (NullPointerException ignore) {
@@ -613,6 +618,8 @@ public class WireMarshaller<T> {
         }
 
         protected void copy(Object from, Object to) throws IllegalAccessException {
+            triggerFinalWarning();
+
             ObjectUtils.requireNonNull(from);
             ObjectUtils.requireNonNull(to);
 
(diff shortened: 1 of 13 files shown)


src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -461,6 +462,10 @@ public boolean isLeaf() {
             offset = unsafeObjectFieldOffset(field);
             key = field::getName;
             this.isLeaf = isLeaf;
+
+            if ((field.getModifiers() & Modifier.FINAL) != 0)
```
minborg, line 466: Suggest: `Modifier.isFinal(field.getModifiers))`

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -450,6 +450,7 @@ public boolean isLeaf() {
 
         Comment commentAnnotation;
         Boolean isLeaf;
+        boolean isFinalNoWarning;
```
minborg, line 453: `final`

minborg: Ahh. ok. You change it after one warning. So no `final`.

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
alamar: We seem to always use the form I've used, can we stick with it for consistency?

minborg: Yes

alamar:
Turns out we have more than 10 tests which do exactly that - overwrite final fields.

What should we do - should we update them all?

@rogersimmons @RobAustin

JerryShea:
> What should we do - should we update them all?
> 
IMO yes we should - our code should set a good example

rogersimmons:
+1. Yes, please fix any tests which overwrite final fields.

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -641,6 +650,14 @@ protected void readValue(Object o, Object defaults, ValueIn read, boolean overwr
             }
         }
 
+        protected void triggerFinalWarning() {
```
minborg, line 653: Suggest `assertNonFinal()`. In this version, there will be a warning but in later, an Exception would be thrown.

src/test/java/net/openhft/chronicle/wire/NestedMapsTest.java
```diff
@@ -255,9 +255,9 @@ public void testMapReadAndWrite() {
     }
 
     static class Mapped extends SelfDescribingMarshallable {
-        final Set<String> words = new LinkedHashSet<>();
-        final List<Integer> numbers = new ArrayList<>();
-        final Map<String, String> map1 = new LinkedHashMap<>();
-        final Map<String, Double> map2 = new LinkedHashMap<>();
+        Set<String> words = new LinkedHashSet<>();
```
minborg, line 258: Can't mutable objects be final?

peter-lawrey:
It's important to recognise how Java Serialization does this currently, as we can't change that behaviour. Java Serialization doesn't produce a warning when setting final fields, nor does it expect you to make a field that can be set non-final

PR #421, opened 2022-03-23
Fixed TEXT parsing of Queue configs.

alamar:


src/main/java/net/openhft/chronicle/wire/YamlWire.java
```diff
@@ -383,18 +383,19 @@ private StringBuilder acquireStringBuilder() {
...
+                sb.setLength(0);
+                sb.append(text);
                 unescape(sb, yt.blockQuote());
                 yt.next();
             } else {
                 throw new IllegalStateException(yt.toString());
             }
         } else {
-            return sb;
+            sb.setLength(0);
```
minborg, line 398: sb.setLength() could be moved to L387 and deduplicate.

minborg (approved):

src/main/java/net/openhft/chronicle/wire/YamlWire.java
alamar: It needs to be called after yt.text()

alamar:
@minborg please re-review as I have made some important changes.

minborg (approved):

Issue #424, opened 2022-04-01
WireMarshaller silently ignores a missing alias


nicktindall:
This is the root cause of https://github.com/ChronicleEnterprise/Chronicle-FIX/issues/966

This exception, that occurs because of a missing alias
```
java.lang.IllegalArgumentException: Cannot find a class for MissingAlias are you missing an alias?
	at net.openhft.chronicle.wire.Wires.tupleFor(Wires.java:619)
	at net.openhft.chronicle.wire.TextWire$TextValueIn.typePrefixOrObject(TextWire.java:3167)
	at net.openhft.chronicle.wire.JSONWire$JSONValueIn.typePrefixOrObject(JSONWire.java:419)
	at net.openhft.chronicle.wire.Wires.object0(Wires.java:526)
	at net.openhft.chronicle.wire.ValueIn.object(ValueIn.java:529)
	at net.openhft.chronicle.wire.JSONWire$JSONValueIn.object(JSONWire.java:408)
	at net.openhft.chronicle.wire.WireMarshaller$ObjectFieldAccess.setValue(WireMarshaller.java:748)
	at net.openhft.chronicle.wire.WireMarshaller$FieldAccess.readValue(WireMarshaller.java:632)
	at net.openhft.chronicle.wire.WireMarshaller.readMarshallableDTOOrder(WireMarshaller.java:303)
	at net.openhft.chronicle.wire.WireMarshaller.readMarshallable(WireMarshaller.java:296)
	at net.openhft.chronicle.wire.Wires.readMarshallable(Wires.java:355)
	at net.openhft.chronicle.wire.SerializationStrategies$6.readUsing(SerializationStrategies.java:185)
	at net.openhft.chronicle.wire.TextWire$TextValueIn.marshallable(TextWire.java:3266)
	at net.openhft.chronicle.wire.JSONWire$JSONValueIn.marshallable(JSONWire.java:436)
	at net.openhft.chronicle.wire.Wires.objectMap(Wires.java:493)
	at net.openhft.chronicle.wire.Wires.object0(Wires.java:574)
	at net.openhft.chronicle.wire.ValueIn.object(ValueIn.java:518)
	at net.openhft.chronicle.wire.JSONWire$JSONValueIn.object(JSONWire.java:403)
```
[...]

PR #425, opened 2022-04-01
Warn reliably when we encounter an unresolvable class

nicktindall:
This is a relatively minimal fix for for #424. I was wary of changing behaviour too much so opted to stop at making the behaviour a bit more consistent when an explicit class name can't be resolved.

I also added the ability to expect vs ignore exceptions consistent with the newer approaches seen in Queue etc. this allows us to test that warnings are being logged when these errors are encountered.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
@@ -3437,11 +3437,10 @@ public class BinaryWire extends AbstractWire implements Wire {
             try {
                 return sb == null ? null : classLookup().forName(sb);
             } catch (ClassNotFoundRuntimeException e) {
-                if (Wires.dtoInterface(tClass)) {
-                    if (GENERATE_TUPLES)
-                        return Wires.tupleFor(tClass, sb.toString());
-                    Jvm.warn().on(getClass(), "Unknown class, perhaps you need to define an alias", e);
+                if (Wires.dtoInterface(tClass) && GENERATE_TUPLES) {
+                    return Wires.tupleFor(tClass, sb.toString());
                 }
+                Jvm.warn().on(getClass(), "Unknown class (" + sb + "), perhaps you need to define an alias", e);
                 return null;
             }
         }
@@ -3560,7 +3559,7 @@ public class BinaryWire extends AbstractWire implements Wire {
 
         @Override
         @Nullable
-        public Object marshallable(@Nullable Object object, @NotNull SerializationStrategy strategy)
+        public Object marshallable(@NotNull Object object, @NotNull SerializationStrategy strategy)
                 throws BufferUnderflowException, IORuntimeException {
             if (this.isNull())
                 return null;
--- a/src/main/java/net/openhft/chronicle/wire/TextWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/TextWire.java
@@ -3136,14 +3136,13 @@ public class TextWire extends AbstractWire implements Wire {
                 bytes.readSkip(-1);
                 try {
                     return classLookup().forName(stringBuilder);
-                } catch (NoClassDefFoundError e) {
-                    throw new IORuntimeException("Unable to load class " + e, e);
                 } catch (ClassNotFoundRuntimeException e) {
                     if (tClass == null) {
                         if (Wires.GENERATE_TUPLES) {
                             return Wires.tupleFor(null, stringBuilder.toString());
                         }
-                        throw new NoClassDefFoundError("Unable to load " + stringBuilder + ", is a class alias missing.");
+                        Jvm.warn().on(TextWire.class, "Unable to load " + stringBuilder + ", is a class alias missing.");
+                        return null;
                     }
 
                     final String className = tClass.getName();
--- a/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
@@ -753,6 +753,7 @@ public class WireMarshaller<T> {
                             !type.isEnum() &&
                             !read.isTyped()) {
                         // retain the null value of object
+                        Jvm.warn().on(getClass(), "Ignoring exception and setting field '" + field.getName() + "' to null", e);
                     } else {
                         Jvm.rethrow(e);
                     }
--- a/src/main/java/net/openhft/chronicle/wire/Wires.java
+++ b/src/main/java/net/openhft/chronicle/wire/Wires.java
@@ -473,25 +473,23 @@ public enum Wires {
     }
 
     @Nullable
-    public static <E> E objectMap(ValueIn in, @Nullable E using, @Nullable Class clazz, SerializationStrategy<E> strategy) {
+    public static <E> E objectMap(ValueIn in, @Nullable E using, @Nullable Class clazz, @NotNull SerializationStrategy<E> strategy) {
         if (in.isNull())
             return null;
-        boolean nullObject = false;
         if (clazz == Object.class)
             strategy = MAP;
         if (using == null) {
             using = (E) strategy.newInstanceOrNull(clazz);
-            nullObject = using == null;
         }
         if (Throwable.class.isAssignableFrom(clazz))
             return (E) WireInternal.throwable(in, false, (Throwable) using);
 
-        if (using == null && !nullObject)
+        if (using == null)
             throw new IllegalStateException("failed to create instance of clazz=" + clazz + " is it aliased?");
+
         final long position = in.wireIn().bytes().readPosition();
         Object marshallable = in.marshallable(using, strategy);
-        // only do a readResolve if an object was provided.
-        E e = using != null ? readResolve(marshallable) : (E) marshallable;
+        E e = readResolve(marshallable);
         String name = nameOf(e);
         if (name != null) {
             E e2 = (E) EnumCache.of(e.getClass()).valueOf(name);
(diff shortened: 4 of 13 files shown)


src/main/java/net/openhft/chronicle/wire/TextWire.java
```diff
@@ -3143,7 +3144,7 @@ public Object typePrefixOrObject(Class tClass) {
                         if (Wires.GENERATE_TUPLES) {
                             return Wires.tupleFor(null, stringBuilder.toString());
                         }
-                        throw new NoClassDefFoundError("Unable to load " + stringBuilder + ", is a class alias missing.");
+                        throw new MissingClassOrClassAliasException("Unable to load " + stringBuilder + ", is a class alias missing.");
```
nicktindall: I feel like whatever happens we should probably not be throwing `NoClassDefFoundError`

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -753,14 +756,15 @@ protected void setValue(Object o, @NotNull ValueIn read, boolean overwrite) thro
                             !type.isEnum() &&
                             !read.isTyped()) {
                         // retain the null value of object
+                        Jvm.warn().on(getClass(), "Ignoring exception and setting field '" + field.getName() + "' to null", e);
```
nicktindall, line 756: It might be nice to log a warning when we swallow this exception? it seems very broad and made this problem harder to spot

src/main/java/net/openhft/chronicle/wire/internal/MissingClassOrClassAliasException.java
```diff
@@ -0,0 +1,16 @@
+package net.openhft.chronicle.wire.internal;
+
+public class MissingClassOrClassAliasException extends RuntimeException {
```
nicktindall: If we DO want a missing class or alias to break deserialization then we should have a consistent error to represent that. It manifests as `NoClassDefFoundError`/`ClassNotFoundRuntimeException`/`IllegalArgumentException` depending on where it's thrown.

src/main/java/net/openhft/chronicle/wire/TextWire.java
nicktindall: Perhaps the correct behaviour is just to log a warning and return null, a lot of other ValueIns do that.

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
JerryShea: Agree - even if that's the only change we make

src/main/java/net/openhft/chronicle/wire/internal/MissingClassOrClassAliasException.java
JerryShea: extend IllegalStateException?

src/test/java/net/openhft/chronicle/wire/AbstractUntypedFieldTest.java
```diff
@@ -49,6 +51,20 @@ void untypedFieldsShouldBeNull(Function<Bytes<byte[]>, Wire> wireConstruction) {
         assertNull(holder.a);
     }
 
+    @ParameterizedTest
+    @MethodSource("provideWire")
+    void missingAliasExceptionsShouldNotBeSwallowed(Function<Bytes<byte[]>, Wire> wireConstruction) {
```
JerryShea: nice

src/test/java/net/openhft/chronicle/wire/AbstractUntypedFieldTest.java
```diff
@@ -49,6 +51,20 @@ void untypedFieldsShouldBeNull(Function<Bytes<byte[]>, Wire> wireConstruction) {
...
+        final Bytes<byte[]> bytes = Bytes.from("!net.openhft.chronicle.wire.AbstractUntypedFieldShouldBeNull$Holder {\n" +
+                "  a: !MissingAlias {\n" +
+                "  }\n" +
+                "}");
+        final Wire textWire = wireConstruction.apply(bytes);
+
+        assertThrows(MissingClassOrClassAliasException.class, () -> {
+            textWire.getValueIn().object(Holder.class);
```
JerryShea: so current behaviour is for the exception to be silently swallowed?

nicktindall: Yeah but only due to that catch block that I put on [the issue](https://github.com/OpenHFT/Chronicle-Wire/issues/424). I don't think it's intentional.

src/main/java/net/openhft/chronicle/wire/TextWire.java
```diff
@@ -3136,14 +3136,12 @@ public Object typePrefixOrObject(Class tClass) {
                 bytes.readSkip(-1);
                 try {
                     return classLookup().forName(stringBuilder);
-                } catch (NoClassDefFoundError e) {
-                    throw new IORuntimeException("Unable to load class " + e, e);
```
nicktindall, line 3140: I also don't think we should be catching it.. 
> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -629,8 +631,8 @@ protected void readValue(Object o, Object defaults, ValueIn read, boolean overwr
                 long pos = read.wireIn().bytes().readPosition();
                 try {
                     setValue(o, read, overwrite);
-                } catch (UnexpectedFieldHandlingException e) {
-                    Jvm.rethrow(e);
+                } catch (FatalMarshallingException e) {
+                    throw e;
```
nicktindall: This is a bit nasty but we need a way for some exceptions to propagate through this catch-all stuff.

src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -476,18 +478,16 @@ public static int removeMaskedTidFromHeader(final int header) {
...
             using = (E) strategy.newInstanceOrNull(clazz);
-            nullObject = using == null;
         }
         if (Throwable.class.isAssignableFrom(clazz))
             return (E) WireInternal.throwable(in, false, (Throwable) using);
 
-        if (using == null && !nullObject)
-            throw new IllegalStateException("failed to create instance of clazz=" + clazz + " is it aliased?");
+        if (using == null)
+            throw new FatalMarshallingException("failed to create instance of clazz=" + clazz + " is it aliased?");
```
nicktindall: This ended up being unreachable from what I could see (it was effectively `if (using == null && using != null)` if you look at what happened up till this point. It meant we ended up passing in `null` to some @NotNull parameters below.

It became that way here https://github.com/OpenHFT/Chronicle-Wire/commit/4d69ee9e7d5cdb40774307d071fb53a568bbf37d#diff-93e71c0374096bba5af3369c6b8ec64a71bad073b960cbefe8ec30703cf81577R436

src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -522,9 +522,16 @@ public static int removeMaskedTidFromHeader(final int header) {
...
+        Object o;
+        try {
+            o = in.typePrefixOrObject(clazz);
+            if (o != null && !(o instanceof Class)) {
+                return (E) in.marshallable(o, MARSHALLABLE);
+            }
+        } catch (MissingClassOrClassAliasException e) {
+            // If the type is missing, warn and try and try to read anyway
+            Jvm.warn().on(Wires.class, e.getMessage());
+            o = null;
```
nicktindall: In this specific case we need to continue trying to parse it, but at least now we've logged a warning. Continuing to attempt to parse means if e.g. the SerializationStrategy returns a value from newInstanceOrNull we might have a chance at deserializing what follows.

Also we need to keep trying in order to satisfy that an unknown enum reference is parsed as a String (there's a test for that)

src/main/java/net/openhft/chronicle/wire/Wires.java
nicktindall: Another odd thing is a few lines down it calls `ValueIn#marshallable(Object, SerializationStrategy)`... the `BinaryValueIn` implementation of that has the first parameter marked as `@Nullable` but all other implementations have it marked as `@NotNull`. I'm going with the majority here and assuming we don't want to pass this point if `using == null`
It might be worth putting those `@Nullable`/`@NotNull` annotations on the interface to make them more consistent?

src/main/java/net/openhft/chronicle/wire/TextWire.java
peter-lawrey: Might be a better option

src/main/java/net/openhft/chronicle/wire/TextWire.java
peter-lawrey: It appears to me,
The Error should be caught in ClassAliasPool.forName0 or not at all. c.f. the catch(ClassNotFoundException)
I assume this behaviour isn't consistent with other calls to forName().

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
peter-lawrey: I assume FatalMarshallingException extends IllegalStateException?

peter-lawrey (approved):

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
```diff
@@ -3560,7 +3559,7 @@ public boolean isNull() {
 
         @Override
         @Nullable
-        public Object marshallable(@Nullable Object object, @NotNull SerializationStrategy strategy)
+        public Object marshallable(@NotNull Object object, @NotNull SerializationStrategy strategy)
```
nicktindall, line 3562: This is consistent with the other implementations of this, and the places that it's called. I added these annotations to the interface as well for consistency.

src/main/java/net/openhft/chronicle/wire/TextWire.java
nicktindall: Yeah it's the only place it was done.

nicktindall: So this was actually necessary when the case was wrong in the classname on Windows. I've moved the fix up to Core https://github.com/OpenHFT/Chronicle-Core/pull/373.

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
nicktindall: I ended up canning this in favour of a more minimal solution

JerryShea (approved):

PR #428, opened 2022-04-07
automatic java17 profile https://github.com/ChronicleEnterprise/RootP…

JerryShea:
…om/issues/218

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>net.openhft</groupId>
         <artifactId>java-parent-pom</artifactId>
-        <version>1.1.29</version>
+        <version>1.1.31</version>
         <relativePath />
     </parent>
 
@@ -194,7 +194,9 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
-                    <argLine>@{argLine} -Duser.timezone=UTC</argLine>
+                    <systemPropertyVariables>
+                        <user.timezone>UTC</user.timezone>
+                    </systemPropertyVariables>
                     <forkCount>1</forkCount>
                     <reuseForks>true</reuseForks>
                     <runOrder>hourly</runOrder>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -328,14 +331,10 @@
...
-                                --add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
-                                --add-opens=jdk.compiler/com.sun.tools.javac=ALL-UNNAMED
-                                --add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
-                                --add-opens=java.base/java.lang=ALL-UNNAMED
-                                --add-opens=java.base/java.lang.reflect=ALL-UNNAMED
+                                --add-exports java.base/jdk.internal.ref=ALL-UNNAMED --add-exports=java.base/sun.nio.ch=ALL-UNNAMED
+                                --add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED --add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
+                                --add-opens=jdk.compiler/com.sun.tools.javac=ALL-UNNAMED --add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
+                                --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.lang.reflect=ALL-UNNAMED
                                 --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED
```
nicktindall: I bet we could achive this without having to duplicate all this stuff by using https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#systemPropertyVariables

minborg: These are not system properties but command line args to `java`. But perhaps I misunderstood your comment?

nicktindall: They're all added in the root parent pom. The only reason we duplicate them here is to add the -Duser.timezone=UTC parameter to argLine. If we use the system property configuration we don't need to duplicate the argLine as it should be inherited?

pom.xml
```diff
@@ -306,43 +308,6 @@
...
-                                --add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
-                                --add-opens=java.base/java.lang=ALL-UNNAMED
-                                --add-opens=java.base/java.lang.reflect=ALL-UNNAMED
-                                --add-opens=java.base/java.io=ALL-UNNAMED --add-opens=java.base/java.util=ALL-UNNAMED
-                            </argLine>
-                        </configuration>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
```
nicktindall, line 345: Don't need this anymore as we're no longer overriding `argLine`

JerryShea: wow! thanks

nicktindall (approved):

PR #436, opened 2022-04-15
Add @Nullable to ValueIn.object arg

glukos:


src/main/java/net/openhft/chronicle/wire/ValueIn.java
```diff
@@ -513,8 +513,12 @@ default <E extends Enum<E>, T> WireIn asEnum(Class<E> eClass, T t, @NotNull BiCo
         return wireIn();
     }
 
+    /**
+     * @param clazz Expected object type. <code>null</code> can be passed to request proxy marshallable tuple
+     *              if {@link Wires#GENERATE_TUPLES} is enabled.
+     */
     @Nullable
-    default <E> E object(@NotNull Class<E> clazz) {
+    default <E> E object(@Nullable Class<E> clazz) {
```
alamar, line 521: Perhaps net.openhft.chronicle.wire.JSONWire.JSONValueIn#object(java.lang.Class<E>) needs updating also.

glukos: Thanks, fixed

alamar:
Please see SonarCloud - looks like we also need to take care of parseType()

glukos:
@alamar Thanks again, I've missed Sonar report.
Non-null clazz was required by `parseType` to check inheritance - now `null` value is treated like `Object` that inherits everything.

alamar (approved):

Issue #438, opened 2022-04-20
Check a Comparable type can be deserialized as a Scalar before writing it as a Scalar type


peter-lawrey:
c.f. https://github.com/OpenHFT/Chronicle-Queue/issues/1024

PR #439, opened 2022-04-20
Check a Comparable type can be deserialized as a Scalar before writing it as a Scalar type, #438

peter-lawrey:


src/test/java/net/openhft/chronicle/wire/serializable/MonetaTest.java
```diff
@@ -0,0 +1,45 @@
...
+    @Test
+    public void monetary() {
+        SortedSet<NonScalarComparable> set = new TreeSet<>();
+        for (String curr : "GBP,USD,EUR,AUD".split(","))
+            set.add(new NonScalarComparable(Currency.getInstance(curr)));
+
+        Wire wire = new TextWire(Bytes.allocateElasticOnHeap());
+        wire.write("currencies")
+                .object(set);
+//        System.out.println(wire);
```
JerryShea: can we get rid of commented-out code?

src/test/java/net/openhft/chronicle/wire/Issue341Test.java
```diff
@@ -26,8 +26,8 @@ public Issue341Test(WireType wireType) {
     public static Collection<Object[]> data() {
         return Arrays.asList(new Object[][]{
                 new Object[]{WireType.TEXT},
-                new Object[]{WireType.JSON},
-                new Object[]{WireType.BINARY},
+                new Object[]{WireType.YAML},
```
JerryShea, line 35: why change?

src/test/java/net/openhft/chronicle/wire/serializable/MonetaTest.java
peter-lawrey: Removed

src/test/java/net/openhft/chronicle/wire/Issue341Test.java
peter-lawrey: Added JSON_ONLY. Had dropped BINARY as have BINARY_LIGHT

PR #445, opened 2022-04-27
Use UnsafeMemory.unsafe(Get|Set)XXX to get and set fields in FieldInfo, ChronicleEnterprise/Chronicle-FIX#969

nicktindall:
This change switches to using unsafeGet/unsafePut to get and set fields instead of using Field#get and Field#set.

We're still missing some specific get/set methods for some primitives in the FieldInfo interface, these will fall back to the reflection implementation which will box/unbox.

I made the first steps towards internalising all the `FieldInfo` implementations.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/DynamicEnum.java
+++ b/src/main/java/net/openhft/chronicle/wire/DynamicEnum.java
@@ -20,7 +20,7 @@ public interface DynamicEnum extends CoreDynamicEnum, Marshallable {
         E nums = cache.valueOf(e.name());
         List<FieldInfo> fieldInfos = e.$fieldInfos();
         for (FieldInfo fieldInfo : fieldInfos) {
-            fieldInfo.set(nums, fieldInfo.get(e));
+            fieldInfo.copy(e, nums);
         }
     }
 }
--- a/src/main/java/net/openhft/chronicle/wire/FieldInfo.java
+++ b/src/main/java/net/openhft/chronicle/wire/FieldInfo.java
@@ -17,12 +17,66 @@
  */
 package net.openhft.chronicle.wire;
 
+import net.openhft.chronicle.wire.internal.fieldinfo.*;
+import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static net.openhft.chronicle.wire.WireMarshaller.WIRE_MARSHALLER_CL;
 
 /**
  * Provides information about a single field of a class or an interface.
  */
 public interface FieldInfo {
+
+    @SuppressWarnings("deprecation" /* VanillaFieldInfo will become internal in the future, remove this suppression in x.25 */)
+    static FieldInfo createForField(String name, Class<?> type, BracketType bracketType, @NotNull Field field) {
+        if (!type.isPrimitive()) {
+            return new ObjectFieldInfo(name, type, bracketType, field);
+        } else if (type == int.class) {
+            return new IntFieldInfo(name, type, bracketType, field);
+        } else if (type == double.class) {
+            return new DoubleFieldInfo(name, type, bracketType, field);
+        } else if (type == long.class) {
+            return new LongFieldInfo(name, type, bracketType, field);
+        } else if (type == char.class) {
+            return new CharFieldInfo(name, type, bracketType, field);
+        }
+        return new VanillaFieldInfo(name, type, bracketType, field);
+    }
+
+    @NotNull
+    static Wires.FieldInfoPair lookupClass(@NotNull Class<?> aClass) {
+        final SerializationStrategy<?> ss = Wires.CLASS_STRATEGY.get(aClass);
+        switch (ss.bracketType()) {
+            case NONE:
+            case SEQ:
+                return Wires.FieldInfoPair.EMPTY;
+            case MAP:
+                break;
+            default:
+                // assume it could be a map
+                break;
+        }
+
+        @NotNull List<FieldInfo> fields = new ArrayList<>();
+        final WireMarshaller<?> marshaller = WIRE_MARSHALLER_CL.get(aClass);
+        for (@NotNull WireMarshaller.FieldAccess fa : marshaller.fields) {
+            final String name = fa.field.getName();
+            final Class<?> type = fa.field.getType();
+            final SerializationStrategy<?> ss2 = Wires.CLASS_STRATEGY.get(type);
+            final BracketType bracketType = ss2.bracketType();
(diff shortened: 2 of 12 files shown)


src/test/java/net/openhft/chronicle/wire/marshallable/FieldInfoTest.java
```diff
@@ -244,6 +244,12 @@ public static Collection<Object[]> combinations() {
                         "  bracketType: NONE,\n" +
                         "  parent: !type net.openhft.chronicle.wire.marshallable.ScalarValues\n" +
                         "}\n" +
+                        ", !FieldInfo {\n" +
+                        "  name: dynamicEnum,\n" +
+                        "  type: !type net.openhft.chronicle.wire.marshallable.TestDynamicEnum,\n" +
+                        "  bracketType: UNKNOWN,\n" +
+                        "  parent: !type net.openhft.chronicle.wire.marshallable.ScalarValues\n" +
+                        "}\n" +
```
nicktindall, line 252: This should use `VanillaFieldInfo`, because `DynamicEnumFieldInfo` is only for use on **fields of** dynamic enums

src/main/java/net/openhft/chronicle/wire/VanillaFieldInfo.java
```diff
@@ -18,111 +18,94 @@
...
-            final String name = fa.field.getName();
-            final Class<?> type = fa.field.getType();
-            final SerializationStrategy ss2 = Wires.CLASS_STRATEGY.get(type);
-            final BracketType bracketType = ss2.bracketType();
-            fields.add(new VanillaFieldInfo(name, type, bracketType, fa.field));
-        }
-        return new Wires.FieldInfoPair(
-                Collections.unmodifiableList(fields),
-                fields.stream().collect(Collectors.toMap(FieldInfo::name, f -> f)));
+        return FieldInfo.lookupClass(aClass);
```
nicktindall, line 52: It seemed like the right thing to do to move this to `FieldInfo` then we can make `VanillaFieldInfo` internal in x.25

src/main/java/net/openhft/chronicle/wire/VanillaFieldInfo.java
```diff
@@ -18,111 +18,98 @@
...
+ */
+@Deprecated
 @SuppressWarnings("rawtypes")
 public class VanillaFieldInfo extends AbstractFieldInfo implements FieldInfo {
 
+    private static final long UNSET_OFFSET = Long.MAX_VALUE;
+
     private final Class parent;
     private transient Field field;
+    private transient long offset = UNSET_OFFSET;
```
nicktindall: If we go through a serialize/deserialize cycle it seems prudent to refresh the offset as we may be on different JVM or have a different implementation of the class available on the classpath.

src/main/java/net/openhft/chronicle/wire/VanillaFieldInfo.java
```diff
@@ -18,111 +18,94 @@
...
-        return new Wires.FieldInfoPair(
-                Collections.unmodifiableList(fields),
-                fields.stream().collect(Collectors.toMap(FieldInfo::name, f -> f)));
+        return FieldInfo.lookupClass(aClass);
     }
 
     @Nullable
     @Override
-    public Object get(Object value) {
+    public Object get(Object object) {
```
nicktindall, line 57: Just made these parameters match the names in the interface, also `value` doesn't make as much sense

src/main/java/net/openhft/chronicle/wire/VanillaFieldInfo.java
```diff
@@ -18,112 +18,90 @@
...
         try {
-            return getField().get(value);
-        } catch (@NotNull NoSuchFieldException | IllegalAccessException e) {
-            e.printStackTrace();
```
nicktindall, line 82: There were lots of these `e.printStackTrace()` calls. They are logged to debug on the following line so I deleted them.

src/main/java/net/openhft/chronicle/wire/DynamicEnum.java
```diff
@@ -20,7 +20,7 @@ static <E extends DynamicEnum> void updateEnum(E e) {
         E nums = cache.valueOf(e.name());
         List<FieldInfo> fieldInfos = e.$fieldInfos();
         for (FieldInfo fieldInfo : fieldInfos) {
-            fieldInfo.set(nums, fieldInfo.get(e));
+            fieldInfo.copy(e, nums);
```
nicktindall, line 23: This assumed all fields were objects, which would use the inefficient boxing/unboxing implementation for primitives

src/main/java/net/openhft/chronicle/wire/VanillaFieldInfo.java
JerryShea: That will make sonar happy too

JerryShea (approved):

src/main/java/net/openhft/chronicle/wire/VanillaFieldInfo.java
```diff
@@ -134,7 +117,12 @@ public double getDouble(Object value) {
     public void set(Object object, Object value) throws IllegalArgumentException {
         Object value2 = ObjectUtils.convertTo(type, value);
         try {
-            getField().set(object, value2);
+            if (type.isPrimitive()) {
+                // use reflection as it does unboxing
+                getField().set(object, value);
```
nicktindall: This prevents segfaults if someone erroneously calls set(...) for a primitive field

src/main/java/net/openhft/chronicle/wire/FieldInfo.java
```diff
@@ -145,5 +200,34 @@ public interface FieldInfo {
...
 
+    /**
+     * Copy the field from the source object to the destination
+     * <p>
+     * Note: this is not a deep copy, objects will be copied by reference
+     *
+     * @param source      The object to copy the field value from
+     * @param destination The object to copy the field value to
+     */
+    default void copy(Object source, Object destination) {
```
JerryShea, line 210: do we still need this?

nicktindall: We don't need it, but it's probably nice to have. To copy a field value from one instance to another you can use `set(dest, get(source))`, which will use reflection and box/unbox in the event it's a primitive field, or you can write an if/else if/else if... to use the correct primitive setter/getter, but if we have this method each field type can implement it's own efficient way of doing it.
It's only useful in one place (in Wire) that I can see (dynamic enums), so you could argue it's unnecessary, but it's not a big cost to implement/maintain.

nicktindall: Oh and if you add a method to an interface binary-compatibility-checker makes you add a default implementation too. I just put the box/unbox one in as the default

JerryShea (approved):

PR #448, opened 2022-05-04
removed deprecated eventName parameter, Fix #437

### LINKED ISSUE #437: Address // TODO: x.23 comments

vach-chronicle:


src/main/java/net/openhft/chronicle/wire/Event.java
```diff
@@ -68,15 +68,9 @@ default E eventTimeNow() {
...
-     *
-     * @param eventName name of the event
+     * Updates event with event time to now if required.
      */
-    // TODO: x.23 remove eventName parameter
-    default E updateEvent(final String eventName) {
-        if (this.eventId().length() == 0)
-            this.eventId(eventName);
-
+    default E updateEvent() {
```
JerryShea, line 73: ah actually I don't think we should do this any more. We may undeprecate eventName

JerryShea (changes requested):
to be discussed

src/main/java/net/openhft/chronicle/wire/Event.java
vach-chronicle: great I was feeling unease about this one )

JerryShea:
Let's close this one without merging. I think we have to bump deprecation of eventId to a later release until we better understand customer usage

vach-chronicle:
Ok no problem, I'll keep the local brunch just in case

JerryShea:
Replaced by #463

PR #449, opened 2022-05-04
Adding support for typed JSON Issue/446

peter-lawrey:


RobAustin (approved):

src/main/java/net/openhft/chronicle/wire/ValueOut.java
```diff
@@ -481,7 +484,11 @@ default WireOut typedMarshallable(@Nullable Serializable object) {
     }
 
     default boolean isScalar(Serializable object) {
-        return object instanceof Comparable;
+        if (object instanceof Comparable) {
```
JerryShea, line 487: you need to rebase I think - this change has already been made to develop

JerryShea (changes requested):
needs to be rebased

src/main/java/net/openhft/chronicle/wire/ValueOut.java
peter-lawrey: Rebase performed

PR #454, opened 2022-05-07
Remove delegate fallback to proxy from generated MethodReader to proxied MethodReader

JerryShea:
But unfortunately some tests fail. 

* `net.openhft.chronicle.wire.MethodReaderDelegationTest#testUnsuccessfulCallIsDelegatedTextWire` and friends fails because we don't support `defaultParselet` with generated code
* `net.openhft.chronicle.wire.VanillaWireParserTest#shouldDetermineMethodNamesFromMethodIds` fails because the test itself is broken - see https://github.com/OpenHFT/Chronicle-Wire/pull/456 for a fix
* `net.openhft.chronicle.wire.method.HandleSkippedValueReadsTest#test` - requires more investigation

And in addition:

* generated MethodReaders don't support [ignoreDefaults ](https://github.com/OpenHFT/Chronicle-Wire/blob/3bcca1204a447b3ca07d9745f17b981404a3de53/src/main/java/net/openhft/chronicle/wire/VanillaMethodReaderBuilder.java#L118). We should [deprecate this property](https://github.com/OpenHFT/Chronicle-Wire/pull/457) and discuss whether we need to support in generated code until gone
* generated MethodWriters don't support [handlerSupplier.methodWriterInterceptorReturns()](https://github.com/OpenHFT/Chronicle-Wire/blob/3bcca1204a447b3ca07d9745f17b981404a3de53/src/main/java/net/openhft/chronicle/wire/VanillaMethodWriterBuilder.java#L189). We should [deprecate this property](https://github.com/OpenHFT/Chronicle-Wire/pull/457) and discuss whether we need to support in generated code until gone

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -118,8 +118,6 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-params</artifactId>
-            <!-- Todo: remove explicit version when 3rd party BOM 3.22.2 is released-->
-            <version>5.8.1</version>
             <scope>test</scope>
         </dependency>
 
--- a/src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
+++ b/src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
@@ -28,20 +28,17 @@ import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.Map;
 import java.util.function.Consumer;
-import java.util.function.Supplier;
 
 import static java.lang.ThreadLocal.withInitial;
 import static net.openhft.chronicle.core.io.Closeable.closeQuietly;
 
 /**
  * Base class for generated method readers.
- * In case generated instance fails to perform a read, it's delegated to lazy-initialized {@link VanillaMethodReader}.
  */
 public abstract class AbstractGeneratedMethodReader implements MethodReader {
     private static final Consumer<MessageHistory> NO_OP_MH_CONSUMER = Mocker.ignored(Consumer.class);
     private final MarshallableIn in;
     protected final WireParselet debugLoggingParselet;
-    private final Supplier<MethodReader> delegateSupplier;
 
     protected MessageHistory messageHistory;
     protected boolean dataEventProcessed;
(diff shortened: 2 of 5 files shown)


glukos:
My big +1 for this activity. A quick assessment:

> net.openhft.chronicle.wire.MethodReaderDelegationTest#testUnsuccessfulCallIsDelegatedTextWire and friends fails because we don't support defaultParselet with generated code

Looks like we have to support the default parselet in the generated code. As a bonus, we'll automatically support `VanillaMethodReaderBuilder#warnMissing`.
Currently, if a generated method reader faces something it can't handle, it goes "panic" - rolls back its progress and retuns `false` (asking the delegate to catch up). We can't keep this behavior if we remove a delegate, otherwise we'll break the contract of `readOne()`: it should return `false` to the caller if and only if the reader didn't make any progress in reading underlying stream of data. Reading something unparseable is still a progress - we should ask default parselet to take care of "unknown" event and move on.

> net.openhft.chronicle.wire.method.HandleSkippedValueReadsTest#test - requires more investigation

The test covers meta-data handlers (added by Peter not so long ago in #400), but they are supported in the generated version. The reason why the test fails (I believe) is that it also covers handling unknown events, which should be taken care of by the default parselet.

> generated MethodReaders don't support ignoreDefaults . We should deprecate this property and discuss whether we need to support in generated code until gone

+1 for deprecation - I don't see a usage pattern for this property (probably there were one, but it's gone)

> generated MethodWriters don't support handlerSupplier.methodWriterInterceptorReturns(). We should deprecate this property and discuss whether we need to support in generated code until gone

Are we going to also get rid of proxies for method writes under the scope of this issue? Anyway, +1 for deprecation - it looks like `updateInterceptor` supersedes it

It's also important to ensure that there are no more known issues that may prevent runtime compilation of generated code. I didn't see any compilation errors after merge of https://github.com/OpenHFT/Java-Runtime-Compiler/pull/104, so fingers crossed.

JerryShea:
Thanks for the feedback @glukos !

Regarding scope, I don't want to remove proxies altogether, just turn off the delegation and ensure that the generated code implements what it needs to (e.g. `defaultParselet`). I'll work on this in the background and will ping you in another week or so for another review.

JerryShea:
See also #472

JerryShea:
@glukos I have added `defaultParselet` support and now this all seems to be working!

Issue #159, opened 2020-04-02
clean up MethodWriter interceptors


JerryShea:
Right now we have 

* MethodReaderInterceptor - deprecated
* MethodReaderInterceptorReturns - the latest version 

and for MethodWriter we have

* MethodWriterListener - deprecated as part of this issue
* MethodWriterInterceptor - deprecated
* MethodWriterInterceptorReturns - the latest version

JerryShea:
See also #148 - IMHO we should *not* allow the user to specify a MethodWriterListener as well as a MethodWriterInterceptor/MethodWriterInterceptorReturns. We should only allow one and if they want to chain them then they can.

JerryShea:
We now have UpdateInterceptor too which is currently only implemented for generated code (see MethodWriterProxyTest). Should we also implement default overloads for >1 arguments?

Also comment from @peter-lawrey in https://github.com/OpenHFT/Chronicle-Wire/issues/185

JerryShea:
Have also committed some extra tests (with @Ignores) to show proxy generated MethodWriters vs generated code.

RobAustin:
As discussed I will ensure that the UpdateInterceptor works with our use of the Java Proxies, and deprecate the MethodWriterInterceptor/MethodWriterInterceptorReturns and all code that uses MethodWriterInterceptor/MethodWriterInterceptorReturns should be changed to use the UpdateInterceptor.

RobAustin:
So we will go from this:

* MethodWriterListener - deprecated as part of this issue
* MethodWriterInterceptor - deprecated
* MethodWriterInterceptorReturns - the latest version

to this:

* MethodWriterListener - deprecated as part of this issue
* MethodWriterInterceptor - deprecated
* MethodWriterInterceptorReturns - deprecated 
* UpdateInterceptor - the latest version

hft-team-city:
Released in [Chronicle-Wire-2.20.29](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.20.29), [BOM-2.20.56](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.20.56)

hft-team-city:
Released in [Chronicle-Wire-2.20.101](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.20.101), [BOM-2.20.134](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.20.134)

PR #457, opened 2022-05-07
Deprecate unused/unsupported VanillaMethodWriterBuilder methods

JerryShea:
These are not used in any of our code, and not exercised by any unit tests

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/VanillaMethodWriterBuilder.java
+++ b/src/main/java/net/openhft/chronicle/wire/VanillaMethodWriterBuilder.java
@@ -118,6 +118,7 @@ public class VanillaMethodWriterBuilder<T> implements Builder<T>, MethodWriterBu
         return this;
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/net/openhft/chronicle/wire/VanillaMethodWriterBuilder.java
```diff
@@ -118,6 +118,7 @@ public MethodWriterBuilder<T> addInterface(Class additionalClass) {
         return this;
     }
 
+    @Deprecated(/* Replaced by UpdateInterceptor. To be removed in x.24 */)
```
JerryShea, line 121: I know we normally give two releases warning, but this has not been used internally since https://github.com/OpenHFT/Chronicle-Wire/issues/159. If any customers are using this, it is straightforward to replace with UpdateInterceptor

nicktindall (approved):

glukos (approved):

Issue #453, opened 2022-05-06
Reduce SQ Issues


minborg:
The quality gate is not passed any longer:

<img width="1426" alt="image" src="https://user-images.githubusercontent.com/7457876/167149701-82547f05-3625-40c3-bbdc-e5fbabd62071.png">

JerryShea:
@minborg I think this was always this high, it is just showing up because we have fixed up the Sonar builds. I'll just increase the number now (to 728) as I have a few PRs to get through, and we can ratchet it down.

minborg:
That is good but I think we should reopen the issue.

hft-team-city:
Released in [Chronicle-Wire-2.23ea9](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.23ea9), [BOM-2.23ea44](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.23ea44)

PR #464, opened 2022-05-09
Add some missing generic params, fix some sonar warnings

nicktindall:
I don't _think_ there are any breakers in this... at least binary compatibility checker said there wasn't but happy for the review. Want to get the sonar warnings down so we can go back to having a single quality gate for all projects.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/AbstractClassGenerator.java
+++ b/src/main/java/net/openhft/chronicle/wire/AbstractClassGenerator.java
@@ -16,26 +16,26 @@ import java.lang.reflect.Parameter;
 import java.util.*;
 import java.util.stream.Collectors;
 
-public abstract class AbstractClassGenerator<MD extends AbstractClassGenerator.MetaData<MD>> {
+public abstract class AbstractClassGenerator<M extends AbstractClassGenerator.MetaData<M>> {
     public static final CachedCompiler CACHED_COMPILER = new CachedCompiler(Jvm.isDebug() ? new File(OS.getTarget(), "generated-test-sources") : null, null);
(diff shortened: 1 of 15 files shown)


src/main/java/net/openhft/chronicle/wire/AbstractClassGenerator.java
```diff
@@ -16,26 +16,26 @@
 import java.util.*;
 import java.util.stream.Collectors;
 
-public abstract class AbstractClassGenerator<MD extends AbstractClassGenerator.MetaData<MD>> {
+public abstract class AbstractClassGenerator<M extends AbstractClassGenerator.MetaData<M>> {
```
minborg, line 19: This class could be `protected` as there is only a protected constructor.

src/main/java/net/openhft/chronicle/wire/AbstractMethodWriterInvocationHandler.java
```diff
@@ -42,7 +42,7 @@ public abstract class AbstractMethodWriterInvocationHandler extends AbstractInvo
     private BiConsumer<Method, Object[]> handleInvoke;
     private boolean useMethodIds;
 
-    protected AbstractMethodWriterInvocationHandler(Class tClass) {
+    protected AbstractMethodWriterInvocationHandler(Class<?> tClass) {
```
minborg, line 45: Because this class is in the public API, anyone can extend the class, and if such an extending class exists and overrides a raw parameter it may lead to a problem. I think we should do this change anyhow.

minborg:
Related to https://github.com/OpenHFT/Chronicle-Wire/issues/453

minborg (approved):

minborg:
Keep 'em coming'

PR #466, opened 2022-05-09
Streaming and Reduction basics in domestic API

minborg:
For a detailed description on these features, please see:

https://github.com/OpenHFT/Chronicle-Wire/tree/issue/440/src/main/java/net/openhft/chronicle/wire/domestic/stream

https://github.com/OpenHFT/Chronicle-Wire/tree/issue/440/src/main/java/net/openhft/chronicle/wire/domestic/reduction


src/main/java/net/openhft/chronicle/wire/TextReadDocumentContext.java
```diff
@@ -150,7 +150,7 @@ public void rollbackOnClose() {
 
     @Override
     public long index() {
-        return 0;
+        return readPosition;
```
RobAustin, line 164: why is the index the read position ?

minborg: The index position is something that can allow us to reposition the "cursor" of the source. For a queue, it is the message index and for a `Wire` it is the `readPosition` (the only thing I can come up with).

src/main/java/net/openhft/chronicle/wire/internal/InternalAutoTailers.java
```diff
@@ -0,0 +1,100 @@
...
+        @Override
+        public void run() {
+            try {
+                while (running) {
+                    if (ReductionUtil.accept(tailer, excerptListener) != -1) {
+                        pauser.pause();
+                    }
+                }
+            } finally {
+                closer.run();
```
RobAustin: I guess the `closer` is from the base class, is this not a bit ugly

src/main/java/net/openhft/chronicle/wire/internal/InternalAutoTailers.java
```diff
@@ -0,0 +1,100 @@
...
+                                 @NotNull final ExcerptListener excerptListener) {
+            requireNonNull(tailerSupplier);
+            this.excerptListener = requireNonNull(excerptListener);
+            this.tailer = requireNonNull(tailerSupplier.get());
+            this.closer = closer(tailer);
+        }
+
+        @Override
+        public final void close() {
+            running = false;
```
RobAustin, line 97: should this not call closer.run() on the current thread - as there look to be no guarantees that the closer is called? it relies on action() being called

src/main/java/net/openhft/chronicle/wire/internal/extractor/DocumentExtractorUtil.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public static <I, E>
+    DocumentExtractor<E> ofMethod(@NotNull final Class<I> type,
+                                  @NotNull final BiConsumer<? super I, ? super E> methodReference,
+                                  @Nullable final Supplier<? extends E> supplier) {
+        final MethodNameAndMessageType<E> info = methodOf(type, methodReference);
+        final String expectedEventName = info.name();
+        final Class<E> elementType = info.messageType();
+        final StringBuilder eventName = new StringBuilder();
```
RobAustin, line 29: can we use the StringBuilderPool ?

RobAustin (approved):

src/main/java/net/openhft/chronicle/wire/internal/InternalAutoTailers.java
minborg: I think the Tailer can ony be used by one thread. So, the event handler needs to close it.

src/main/java/net/openhft/chronicle/wire/internal/extractor/DocumentExtractorUtil.java
minborg: The StringBuilder is only created once per  ofMethod() call so it is being reused for all calls on the DocumentExtractor. I think reusing SBs across several DocumentExtractors would be dangerous.

PR #469, opened 2022-05-10
Infinite message loop is possible with dontWriteHistoryForEventsWithNoOutput=false

glukos:


src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
```diff
@@ -44,6 +44,8 @@ public abstract class AbstractGeneratedMethodReader implements MethodReader {
     private final Supplier<MethodReader> delegateSupplier;
 
     protected MessageHistory messageHistory;
+    protected boolean dataEventProcessed;
```
JerryShea, line 47: I thought I saw a test in this PR before?

JerryShea, line 47: what happened to the test? I thought there was a test in this PR?

JerryShea (approved):
Approved

src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
JerryShea: ah - https://github.com/ChronicleEnterprise/Chronicle-Services/pull/331

PR #473, opened 2022-05-13
Feature/method reader dont delegate to proxy

JerryShea:
Replaces #454 which was closed in error - see also comments made there

This PR builds upon the following fixes: #456 #457 #458 #472 

This PR stops generated method readers falling back to proxy method readers. If you get a problem with generated method readers then you can always disable them with `-DdisableReaderProxyCodegen`.

It also ensures that problems with generated method readers/writer compilation are logged at warning level and the user is asked to report the problem, and that when we fall back to proxies we see a log warning the user that support will be dropped.

We should now be able to drop proxy method readers/writers in x.25.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
+++ b/src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
@@ -28,20 +28,17 @@ import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.Map;
 import java.util.function.Consumer;
-import java.util.function.Supplier;
 
 import static java.lang.ThreadLocal.withInitial;
 import static net.openhft.chronicle.core.io.Closeable.closeQuietly;
 
 /**
  * Base class for generated method readers.
- * In case generated instance fails to perform a read, it's delegated to lazy-initialized {@link VanillaMethodReader}.
  */
 public abstract class AbstractGeneratedMethodReader implements MethodReader {
     private static final Consumer<MessageHistory> NO_OP_MH_CONSUMER = Mocker.ignored(Consumer.class);
     private final MarshallableIn in;
     protected final WireParselet debugLoggingParselet;
-    private final Supplier<MethodReader> delegateSupplier;
 
     protected MessageHistory messageHistory;
     protected boolean dataEventProcessed;
@@ -74,11 +71,9 @@ public abstract class AbstractGeneratedMethodReader implements MethodReader {
     private static final MessageHistoryThreadLocal TEMP_MESSAGE_HISTORY = new MessageHistoryThreadLocal();
 
     protected AbstractGeneratedMethodReader(MarshallableIn in,
-                                            WireParselet debugLoggingParselet,
-                                            Supplier<MethodReader> delegateSupplier) {
+                                            WireParselet debugLoggingParselet) {
         this.in = in;
         this.debugLoggingParselet = debugLoggingParselet;
-        this.delegateSupplier = delegateSupplier;
     }
 
     /**
@@ -238,12 +233,9 @@ public abstract class AbstractGeneratedMethodReader implements MethodReader {
             ok = context.isMetaData()
                     ? readOneMeta(context)
                     : readOne0(context);
-
-            if (!ok)
-                context.rollbackOnClose();
         }
 
-        return ok || delegate().readOne();
+        return ok;
     }
 
     public void throwExceptionIfClosed() {
@@ -304,13 +296,6 @@ public abstract class AbstractGeneratedMethodReader implements MethodReader {
             messageHistory = MessageHistory.get();
 
         return messageHistory;
-    }
-
-    private MethodReader delegate() {
-        if (delegate == null)
-            delegate = delegateSupplier.get();
-
-        return delegate;
     }
 
     /**
--- a/src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
+++ b/src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
@@ -159,7 +159,6 @@ public class GenerateMethodReader {
                 "import net.openhft.chronicle.wire.*;\n" +
                 "import net.openhft.chronicle.bytes.MethodReaderInterceptorReturns;\n" +
                 "\n" +
-                "import java.util.function.Supplier;\n" +
                 "import java.util.Map;\n" +
                 "import java.lang.reflect.Method;\n" +
                 "\n");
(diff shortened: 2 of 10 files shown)


src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
```diff
@@ -243,7 +238,7 @@ public boolean readOne() {
                 context.rollbackOnClose();
         }
 
-        return ok || delegate().readOne();
```
JerryShea, line 246: this should improve performance for when there is nothing to read

src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
```diff
@@ -316,10 +319,7 @@ private void generateSourceCode() {
                 "throw e;\n" +
                 "}\n" +
                 "catch (Exception e) {\n" +
-                "Jvm.warn().on(this.getClass(), \"Failure to dispatch message, " +
-                "will retry to process without generated code: \" + lastEventName + \"(), " +
-                "bytes: \" + wireIn.bytes().toDebugString(), e);\n" +
-                "return false;\n" +
+                "throw new InvocationTargetRuntimeException(e);\n" +
```
JerryShea, line 322: I'm not 100% sure about throwing this - maybe we should just not catch this exception?

src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
glukos: Also this should improve performace if we process "unwanted" event, e.g. if a service processes an event that it's not interested in

glukos:
Looks good

glukos (approved):

src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
```diff
@@ -260,20 +263,14 @@ private void generateSourceCode() {
         sourceCode.append(eventNameSwitchBlock);
 
         sourceCode.append("default:\n" +
-                "valueIn.skipValue();\n" +
```
glukos, line 263: I suspect the code above is incorrect now
```
        sourceCode.append("default:\n" +
                "valueIn.skipValue();\n" +
                "return false;\n" +
                "}\n" +
                "}\n" +
                "else {\n" +
                "lastEventName = wireIn.readEvent(String.class);\n" +
                "}\n" +
                "try {\n" +
                "if (Jvm.isDebug())\n" +
                "debugLoggingParselet.accept(lastEventName, valueIn);\n" +
                "if (lastEventName == null)\n" +
                "throw new IllegalStateException(\"Failed to read method name or ID\");\n" +
                "switch (lastEventName) {\n" +
                "case MethodReader.HISTORY:\n" +
                "valueIn.marshallable(messageHistory);\n" +
                "break;\n\n");
```
If we detect `BinaryWireCode.FIELD_NUMBER` and we don't find an appropriate method ID in the switch block, we still return `false` hoping that a delegate will handle this case for us.
Looking at `VanillaMethodReader`, I think we should consume the unexpected event using `WireParser#SKIP_READABLE_BYTES` and return `true`.

src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
```diff
@@ -243,7 +238,7 @@ public boolean readOne() {
                 context.rollbackOnClose();
```
glukos: I'm not sure that this part with `rollbackOnClose` makes sense when we don't have a delegate.

Normally, the generated instance should never return `false`, but if it happens, we should skip an event (ideally, with a warning) and go on. Rolling back will force the method reader to spin on such "erroneous" event.

glukos (changes requested):
Please see my comments above. This is crucial code for the Services - it would be great to cover "unkown method ID" case with a test.
I think it's also possible to cover "returning `false` by generated code shouldn't halt the progress" with a test using `GeneratingMethodReaderInterceptorReturns` hack - I think it's worth it.

src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
JerryShea: Thanks - let me review again (and add some tests)

src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
JerryShea: Nice one - well spotted - thanks! Totally agree with you re: comment below and extra tests - this needs to be bulletproof

JerryShea: Yep - I have now removed the rollbackOnClose. The method reader always returns true if there is some data to be read as there is always a delegate (there is a default one even if you don't specify - see `MethodReaderDelegationTest#testUnsuccessfulCallNoDelegate()`)

src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
JerryShea: See above comment - we always return true if there was data to be read. Either the service processes it, or the defaultParser does

src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
JerryShea: @glukos please re-review!

src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
```diff
@@ -240,8 +243,9 @@ private void generateSourceCode() {
         sourceCode.append(eventIdSwitchBlock);
 
         sourceCode.append("default:\n" +
-                "valueIn.skipValue();\n" +
-                "return false;\n" +
+                // below should be garbage-free if methodId is low. This will now drop through for defaultParselet
+                "lastEventName = Integer.toString(methodId);\n" +
```
glukos, line 247: This is not straightforward, but I guess should work - we'll execute "default" section of method name switch

src/main/java/net/openhft/chronicle/wire/AbstractGeneratedMethodReader.java
glukos: Yes - this is correct. My point was that services now will process "unexpected" events faster, which is a very real-life scenario. In case e.g. we have a service that's everyone listens to (but each listener is interested in different events).

glukos (approved):
Looks good, I believe all corner cases are now covered with tests.

PR #476, opened 2022-05-18
Added a YamlTester interface to simplify running YAML based tests #475

peter-lawrey:


src/main/java/net/openhft/chronicle/wire/utils/YamlTester.java
```diff
@@ -0,0 +1,70 @@
...
+
+import net.openhft.chronicle.wire.TextMethodTester;
+
+import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.util.function.Function;
+
+public interface YamlTester {
+    /**
+     * Test a microservice implemented in a class using in.yaml comparing with out.yaml,
```
alamar: Is microservice applicable here?

peter-lawrey: Potentially, but "component" would be better here. Thank you

RobAustin (approved):

PR #481, opened 2022-05-30
Extend zero writes after end of append context to ensure the complete…

rogersimmons:
Fix for Queue issue https://github.com/OpenHFT/Chronicle-Queue/issues/1099


nicktindall (approved):
Can we please backport this to .22 and perhaps even .21 branch too?

JerryShea (approved):

nicktindall:
Can we add to IncompleteMessageTest to reproduce the issue? It seems to have existed since February and not failed so I assume it doesn't cover this case.

JerryShea:
See also https://github.com/OpenHFT/Chronicle-Wire/pull/496 which is the backport to 22

PR #493, opened 2022-06-10
Rationalise LongConverter with more tests and add IdentifierLongConverter #492

peter-lawrey:


src/main/java/net/openhft/chronicle/wire/Base32IntConverter.java
```diff
@@ -18,64 +18,30 @@
...
 
 import net.openhft.chronicle.core.Jvm;
-import net.openhft.chronicle.core.util.StringUtils;
-
-import java.util.Arrays;
 
 /**
  * Unsigned 32-bit number with encoding to be as disambiguated as possible.
  */
+@Deprecated(/* to remove in x.25 */)
```
JerryShea, line 27: replaced by?

src/main/java/net/openhft/chronicle/wire/Base40IntConverter.java
```diff
@@ -25,10 +25,11 @@
 /**
  * Unsigned 32-bit number.
  */
+@Deprecated(/* to remove in x.25 */)
 public class Base40IntConverter implements IntConverter {
     public static final int MAX_LENGTH = IntConverter.maxParseLength(40);
     public static final Base40IntConverter INSTANCE = new Base40IntConverter();
-    private static final String CHARS = ".ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789:-+";
```
JerryShea, line 31: this is a breaking change? i.e. encoded ints will not be compatible before/after? If so, we should either be able to read the old style identifiers or at least warn the user if we encounter an id with a ":"?

peter-lawrey: What isn't clear in this is that using colon meant the value couldn't be read safely as it was (in TextWire or YamlWire). This only worked if you had a value that didn't contain that character.

JerryShea (approved):

src/main/java/net/openhft/chronicle/wire/LongConverter.java
```diff
@@ -63,8 +96,17 @@ default void lengthCheck(CharSequence text) {
             throw new IllegalArgumentException(format("text={0} exceeds the maximum allowable length of {1}", text, maxParseLength()));
     }
 
-    static int maxParseLength(int based) {
-        return (int) (64 / log(based) * log(2));
+    default boolean allSafeChars() {
+        return true;
```
nicktindall, line 106: "safe" is very contextual. I assume this means safe for YAML strings from your explanation, perhaps this should be named more specifically (e.g. `safeForYamlString`). (e.g. XML has a separate set of characters that need to be escaped, JSON different ones and windows/linux filenames different again).

Either that or it could be generic and we could provide an enum like `allSafeChars(ForPurpose.WINDOWS_FILENAME)` ? 

(the enum would implement some interface that took a list of characters and returned true/false depending if the character set was safe for that purpose)

pom.xml
```diff
@@ -188,7 +188,7 @@
                         <configuration>
                             <referenceVersion>2.23ea0</referenceVersion>
                             <artifactsURI>https://teamcity.chronicle.software/repository/download</artifactsURI>
-                            <binaryCompatibilityPercentageRequired>96</binaryCompatibilityPercentageRequired>
+                            <binaryCompatibilityPercentageRequired>95</binaryCompatibilityPercentageRequired>
```
RobAustin, line 191: it's not good to change this.

RobAustin (approved):

src/main/java/net/openhft/chronicle/wire/LongConverter.java
peter-lawrey: I have added a parameter so the type of the WireOut can be used to determine whether they are safe or not.

pom.xml
peter-lawrey: This comes from making some LongConverters use inheritance.

src/main/java/net/openhft/chronicle/wire/AbstractLongConverter.java
```diff
@@ -0,0 +1,35 @@
+package net.openhft.chronicle.wire;
+
+import net.openhft.chronicle.bytes.Bytes;
+
+public abstract class AbstractLongConverter implements LongConverter {
```
minborg, line 5: As discussed, we should not have abstract classes in the public API.

PR #496, opened 2022-06-12
Extend zero writes after end of append context to ensure the complete…

JerryShea:
… next header (int + any padding offset) is guaranteed to be 0 (NONE/EOF) (#481)

Co-authored-by: rogersimmons <roger.simmons@higherfrequencytrading.com>


JerryShea:
Also bring in some changes from develop to make the build work on all platforms

rogersimmons (approved):

PR #503, opened 2022-06-15
Feature/merge ea to develop

JerryShea:


jlbh-test.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+mvn clean  test-compile exec:java -Dexec.mainClass="net.openhft.chronicle.wire.TriviallyCopyableJLBH" -Dexec.classpathScope=test
```
JerryShea, line 2: @j4sm1ne96 @RobAustin please follow the convention of adding to exec plugin and enabling with `run-benchmarks` profile then we get charts in TC for free

nicktindall (approved):

PR #512, opened 2022-06-20
Deprecate AbstractCommonMarshallable in preparation to move it package-private

### LINKED ISSUE #507: AbstractCommonMarshallable deserialisation behaviour change
Customer had a DTO which descended from ACM and deserialisation behaviour changed from 21 to 22.

Consider also deprecating/package privating/internalising.

Consider logging a warning in deserialisation code if ACM used.

nicktindall:
Fixes #507

We have seen people extending this directly in the wild, some time between BOM 2.20 and 2.22 it flipped from being self describing to being not. It's probably better that descenders make that choice explicitly by descending from SelfDescribingMarshallable or BytesInBinaryMarshallable than relying on the default here.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/AbstractCommonMarshallable.java
+++ b/src/main/java/net/openhft/chronicle/wire/AbstractCommonMarshallable.java
@@ -21,7 +21,10 @@ import net.openhft.chronicle.bytes.BytesMarshallable;
 
 /**
  * This uses bytes marshallable, non self describing messages by default.
+ *
+ * @deprecated This will become package-private in x.25 use {@link SelfDescribingMarshallable} or {@link BytesInBinaryMarshallable} instead
  */
+@Deprecated(/* to be made package-private in x.25 */)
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/wire/AbstractCommonMarshallable.java
```diff
@@ -21,7 +21,10 @@
 
 /**
  * This uses bytes marshallable, non self describing messages by default.
+ *
+ * @deprecated This will become package-private in x.24 use {@link SelfDescribingMarshallable} or {@link BytesInBinaryMarshallable} instead
  */
+@Deprecated(/* to be made package-private in x.24 */)
 public abstract class AbstractCommonMarshallable implements Marshallable, BytesMarshallable {
```
JerryShea, line 28: I like this very much but we will need to socialise more broadly within the team as we have generated code "in the wild" that uses this.

JerryShea (approved):

PR #527, opened 2022-06-27
Un-ignore tests in YamlWireTest - closes #510

### LINKED ISSUE #510: Fix @Ignore on YamlWireTest ( in Chronicle-Wire )

alamar:


src/test/java/net/openhft/chronicle/wire/YamlWireTest.java
```diff
@@ -1552,7 +1547,7 @@ public void testByteArray() {
                         "# position: 62, header: 2\n" +
                         "--- !!data\n" +
                         "four: !byte[] !!binary AQIDBA==\n"
-                , Wires.fromSizePrefixedBlobs(wire.bytes()));
```
alamar, line 1555: @RobAustin a recent edit that you have made makes the one-arg version to assume padding=true by default,
but two-arg function assumes padding=false, and that seems to be library-wide default
Should we harmonize it?

RobAustin: I don't remember making this change, I'm not sure if we should harmonize best speak to @peter-lawrey

RobAustin (approved):

src/test/java/net/openhft/chronicle/wire/YamlWireTest.java
peter-lawrey: I agree we need to carefully improve the inconsistencies in how padding is handled.

PR #543, opened 2022-07-20
Json tests

peter-lawrey:


peter-lawrey:
Merged manually, only one small change kept.

PR #544, opened 2022-07-25
Add Syncable interface to mark Syncable resources https://github.com/…

peter-lawrey:
…OpenHFT/Chronicle-Core/issues/433

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/GenerateMethodWriter.java
+++ b/src/main/java/net/openhft/chronicle/wire/GenerateMethodWriter.java
@@ -5,6 +5,7 @@ import net.openhft.chronicle.bytes.MethodReader;
 import net.openhft.chronicle.bytes.UpdateInterceptor;
 import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.io.Closeable;
+import net.openhft.chronicle.core.io.Syncable;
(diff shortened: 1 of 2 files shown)


peter-lawrey:
All builds have actually passed despite what it says here

src/main/java/net/openhft/chronicle/wire/GenerateMethodWriter.java
```diff
@@ -459,14 +460,18 @@ private CharSequence createMethod(SortedSet<String> importSet, final Method dm,
                     ".update(\"" + dm.getName() + "\", " + name + ")) return" + returnDefault(returnType) + ";\n");
         }
 
+        body.append("MarshallableOut out = this.out.get();\n");
+        if (dm.getDeclaringClass() == Syncable.class) {
+            body.append(Syncable.class.getName()).append(".syncIfAvailable(out);\n");
```
JerryShea, line 465: what does this do different?

peter-lawrey: If the interface for the method writer extends Syncable and you call sync() it will call the same on the underlying MarshallableOut if it supports Syncable before writing the `sync: ""`
This way, the caller can determine programmatically when sync() should be performed and the reader can determine at what point sync() was called upstream.
Forough has an issue to create a test in services with a demo.

PR #546, opened 2022-07-28
Added Channels for remote access to gateways #545

### LINKED ISSUE #545: Added Channels for remote access to gateways

peter-lawrey:
Fixes #545

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,6 @@
         <dependency>
             <groupId>net.openhft</groupId>
             <artifactId>compiler</artifactId>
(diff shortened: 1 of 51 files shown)


pom.xml
```diff
@@ -72,6 +72,7 @@
         <dependency>
             <groupId>net.openhft</groupId>
             <artifactId>compiler</artifactId>
+            <version>2.21ea82</version>
```
JerryShea: shouldn't this be in bom?

peter-lawrey: Yes, fixed now

PR #548, opened 2022-08-05
#547 Ensure objects that extends AbstractMarshallableCfg are reset wh…

RobAustin:
…en calling valueIn.object(Object, Class)} #547

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/Wires.java
+++ b/src/main/java/net/openhft/chronicle/wire/Wires.java
@@ -588,6 +588,9 @@ public enum Wires {
 
     @Nullable
     public static <E> E object0(ValueIn in, @Nullable E using, @Nullable Class clazz) {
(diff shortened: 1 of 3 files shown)


maxim-ponomarev (approved):

src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -588,6 +588,9 @@ public static <E> E objectDate(ValueIn in, @Nullable E using) {
 
     @Nullable
     public static <E> E object0(ValueIn in, @Nullable E using, @Nullable Class clazz) {
+        if (using instanceof AbstractMarshallableCfg)
```
JerryShea, line 591: We already have `Marshallable#reset` - isn't this being called? https://github.com/OpenHFT/Chronicle-Wire#a-note-on-wires-reset

PR #552, opened 2022-08-16
Avoid unnecessary CAS operation when writing EOF marker

nicktindall:
Fixes OpenHFT/Chronicle-Queue#1141

The theory is here, the CAS operation causes a mapped page to be changed to read/write which triggers an update to mtime regardless of whether the CAS results in an update. It meant in queue, when `normaliseEOFs` was called on the creation of a new appender, all the roll cycles' mtime was updated to now, even if no change was made to the file.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/AbstractWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/AbstractWire.java
@@ -438,7 +438,7 @@ public abstract class AbstractWire implements Wire {
                     pos += BytesUtil.padOffset(pos);
 
                 if (MEMORY.safeAlignedInt(pos)) {
-                    if (bytes.compareAndSwapInt(pos, 0, END_OF_DATA)) {
+                    if (bytes.readVolatileInt(pos) == 0 && bytes.compareAndSwapInt(pos, 0, END_OF_DATA)) {
                         bytes.writePosition(pos + SPB_HEADER_SIZE);
                         write("EOF");
                         return true;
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/wire/AbstractWire.java
```diff
@@ -438,7 +438,7 @@ public boolean writeEndOfWire(long timeout, TimeUnit timeUnit, long lastPosition
                     pos += BytesUtil.padOffset(pos);
 
                 if (MEMORY.safeAlignedInt(pos)) {
-                    if (bytes.compareAndSwapInt(pos, 0, END_OF_DATA)) {
+                    if (bytes.readVolatileInt(pos) == 0 && bytes.compareAndSwapInt(pos, 0, END_OF_DATA)) {
```
nicktindall, line 441: This works on the assumption that once `END_OF_DATA` is written, it isn't changed to something else. If it is, the `readVolatileInt` might evaluate to false, where a subsequent `compareAndSwap` might have resulted in a change.

src/test/java/net/openhft/chronicle/wire/BinaryWireTest.java
```diff
@@ -1448,6 +1455,29 @@ public void readsComment() {
...
 
+    @Test
+    public void writeEndOfWireDoesNotUpdateModifiedTimeOnNoOpWhenUnderlyingBytesIsFile() throws IOException {
+        final File tempFile = IOTools.createTempFile("test-lastModified-endOfWire");
+        VanillaReferenceOwner owner = new VanillaReferenceOwner("test");
+        final long endOfWirePosition;
+        try (MappedFile mappedFile = SingleMappedFile.mappedFile(tempFile, 10_240)) {
+            final Bytes<?> bytes = mappedFile.acquireBytesForWrite(owner, 0);
+            Wire wire = WireType.BINARY.apply(bytes);
+            wire.write("testing-testing").int8(123);
```
JerryShea, line 1465: :)

JerryShea (approved):

PR #576, opened 2022-09-23
Remove deprecated APIs in 2.24

### LINKED ISSUE #307: HashWire is non-binary (text) wire which leads to object allocation, CPU usage
We have `HashWire` used to calculate hash codes of `SelfDescribingMarshallable`, which is a text wire (it does not override `isBinary()` to return true). Which leads to easy 6-7x CPU usage as well as object allocations. 

I guess we can't change it since it will affect existing objects' hash codes, which could be stored by our users. Maybe we can introduce a system property to make it use the binary form, which we will then make the default (in 2.23 or 3.X?)

alamar:
Remove padding selection and start/end methods
Switch HashWire to binary mode - closes #307

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireInternal.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireInternal.java
@@ -118,41 +118,36 @@ public enum WireInternal {
     public static long writeData(@NotNull WireOut wireOut, boolean metaData, boolean notComplete,
                                  @NotNull WriteMarshallable writer) {
         wireOut.getValueOut().resetBetweenDocuments();
-        assert wireOut.startUse();
         long position;
-        try {
-            @NotNull Bytes<?> bytes = wireOut.bytes();
-            position = bytes.writePositionForHeader(wireOut.usePadding());
 
-            int metaDataBit = metaData ? Wires.META_DATA : 0;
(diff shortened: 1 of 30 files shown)


src/main/java/net/openhft/chronicle/wire/WireInternal.java
```diff
@@ -118,42 +118,29 @@ static StringBuilder acquireAnotherStringBuilder(CharSequence cs) {
...
+        int len0 = metaDataBit | Wires.NOT_COMPLETE | Wires.UNKNOWN_LENGTH;
+        bytes.writeOrderedInt(len0);
+        writer.writeMarshallable(wireOut);
+        if (!wireOut.isBinary())
+            BytesUtil.combineDoubleNewline(bytes);
+        long position1 = bytes.writePosition();
+
+        int length;
+        if (bytes instanceof HexDumpBytes) {
+            // Todo: this looks suspicious. Why cast to int individually rather than use long arithmetics?
```
JerryShea, line 142: agree

alamar: I have no idea, I've just reindented this code

RobAustin (approved):

PR #589, opened 2022-10-28
Implement checkstyle checking for JUnit classnames

nicktindall:
To prevent ChronicleSoftware/RootPom#284

I basically commented out ALL the checkstyle rules because we need to work out an approach to start enabling them. We should probably look at enabling them slowly and perhaps enforcing that the count doesn't increase... There are some good ones in there.

Might be worth considering https://findbugs.sourceforge.net/ as well

This is really just a POC, we should probably put that plugin in the `java-parent-pom`

See the links to the broken builds below

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -435,6 +435,45 @@
                 </executions>
             </plugin>
 
+            <!-- This would end up in java-parent-pom or something -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <version>3.2.0</version>
+                <executions>
+                    <execution>
+                        <id>validate</id>
+                        <phase>validate</phase>
+                        <goals>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -435,6 +435,45 @@
...
+                        See https://checkstyle.sourceforge.io/#JRE_and_JDK
+                    -->
+                    <dependency>
+                        <groupId>com.puppycrawl.tools</groupId>
+                        <artifactId>checkstyle</artifactId>
+                        <version>9.3</version>
+                    </dependency>
+                    <dependency>
+                        <groupId>net.openhft</groupId>
+                        <artifactId>chronicle-quality-rules</artifactId>
```
JerryShea, line 471: cool

JerryShea (approved):

nicktindall:
Will implement this in the java-parent-pom

PR #597, opened 2022-11-17
fix #596 - replace image with new style

### LINKED ISSUE #596: Update Figure 1 in the readme to use the new style

j4sm1ne96:


foroughgoudarzi:
Please use the same color (green) for the len field in the buffer.(only change the color of the len field to green)
the tailer arrow should point to the buffer exactly before the len field.
the Application 2 arrow should point to the buffer exactly after the actual data field.
the Application 1 arrow should point to the beginning of the len field
make the message size curly brace bigger so that it covers the whole length of the len+actual data fields
+
upload the source file of the image

foroughgoudarzi (approved):

PR #600, opened 2022-11-23
Avoid lambda allocations in VanillaMessageHistory

glukos:
I don't know a better way to work this around.
Even JVM 17 fails to handle non-static method references without allocation under some conditions
image

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
+++ b/src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
@@ -24,6 +24,8 @@ import net.openhft.chronicle.core.io.IORuntimeException;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.function.BiConsumer;
+
 @SuppressWarnings("rawtypes")
 public class VanillaMessageHistory extends SelfDescribingMarshallable implements MessageHistory {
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
```diff
@@ -40,6 +42,11 @@ public class VanillaMessageHistory extends SelfDescribingMarshallable implements
     private final long[] sourceIndexArray = new long[MESSAGE_HISTORY_LENGTH];
     @NotNull
     private final long[] timingsArray = new long[MESSAGE_HISTORY_LENGTH * 2];
+
+    // To avoid lambda allocations
+    private final transient BiConsumer<VanillaMessageHistory, ValueOut> acceptSourcesConsumer = this::acceptSources;
```
alamar, line 47: Should we re-set them when class is unmarshalled?

glukos: No - wire marshaller ignores transient fields.
They are set once on construction and don't get replaced / nullified on unmarshalling. Double-checked in `MessageHistoryTest#testReadMarshallable`

alamar: OK

alamar (approved):

PR #605, opened 2022-11-29
Allow checking for EOF marker without writing - needed for …

alamar:
…ChronicleEnterprise/Chronicle-Queue-Enterprise#383

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/AbstractWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/AbstractWire.java
@@ -428,11 +428,11 @@ public abstract class AbstractWire implements Wire {
 
     @Override
     public boolean writeEndOfWire(long timeout, TimeUnit timeUnit, long lastPosition) {
-        return noEndOfWire(true, timeout, timeUnit, lastPosition);
+        return endOfWire(true, timeout, timeUnit, lastPosition) == EndOfWire.PRESENT_AFTER_UPDATE;
     }
 
     @Override
-    public boolean noEndOfWire(boolean writeEOF, long timeout, TimeUnit timeUnit, long lastPosition) {
+    public EndOfWire endOfWire(boolean writeEOF, long timeout, TimeUnit timeUnit, long lastPosition) {
 
         long pos = Math.max(lastPosition, bytes.writePosition());
         headerNumber = Long.MIN_VALUE;
@@ -445,12 +445,12 @@ public abstract class AbstractWire implements Wire {
                 if (MEMORY.safeAlignedInt(pos)) {
                     if (bytes.readVolatileInt(pos) == 0) {
                         if (!writeEOF)
-                            return true;
+                            return EndOfWire.NOT_PRESENT;
 
                         if (bytes.compareAndSwapInt(pos, 0, END_OF_DATA)) {
                             bytes.writePosition(pos + SPB_HEADER_SIZE);
                             write("EOF");
-                            return true;
+                            return EndOfWire.PRESENT_AFTER_UPDATE;
                         }
                     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/wire/AbstractWire.java
```diff
@@ -438,19 +443,26 @@ public boolean writeEndOfWire(long timeout, TimeUnit timeUnit, long lastPosition
...
                     }
 
                 } else {
                     // mis-aligned check, assume only one writer (best effort)
                     MEMORY.loadFence();
                     if (bytes.readInt(pos) == 0) {
-                        bytes.writeInt(pos, END_OF_DATA);
-                        MEMORY.storeFence();
-                        bytes.writePosition(pos + SPB_HEADER_SIZE);
+                        if (writeEOF) {
```
JerryShea, line 461: good catch

JerryShea (approved):

src/main/java/net/openhft/chronicle/wire/WireOut.java
```diff
@@ -193,10 +193,25 @@ default void writeNotCompleteDocument(boolean metaData, @NotNull WriteMarshallab
...
+     * This will increment the headerNumber as appropriate if successful
+     *
+     * @param writeEOF     if {@code true}, write end of wire marker unless already exists
+     * @param timeout      throw TimeoutException if it could not write the marker in time.
+     * @param timeUnit     of the timeout
+     * @param lastPosition the end of the wire
+     * @return {@code true} if did this method could write an EOF (depending on write parameter).
+     *         {@code false} if it was it already there.
+     */
+    default boolean noEndOfWire(boolean writeEOF, long timeout, TimeUnit timeUnit, long lastPosition) {
```
nicktindall: what about if we called it `isEndOfWirePresent` and the `writeEOF` parameter could be called `writeIfNotPresent`

nicktindall: ohh the return value is weird. I guess to support writeEOF

nicktindall: I don't like the method signature, it requires a lot of mental gymnastics to understand.

nicktindall: Perhaps we should return an enum: e.g. `(NotPresent|PresentAfterUpdate|Present)`

nicktindall (changes requested):
We have to change that method signature it's incomprehensible

src/main/java/net/openhft/chronicle/wire/WireOut.java
nicktindall: So to clarify... we return false if the EOF is already there, and true if it's not there. But if `writeEOF` was true and we returned true it means we wrote EOF.

I think it would be clearer if the method were called `isEndOfWirePresent(boolean writeEOFIfNotPresent, ...)`
And it returned `(NotPresent|Present|PresentAfterUpdate)`

i.e.
|State of EOF|writeEOF|existing return value|proposed return value|
|------------------|-------------|------------------------------|---------------------------------|
|present|true|false|Present|
|present|false|false|Present|
|not present|true|true|PresentAfterUpdate|
|not present|false|true|NotPresent|

nicktindall: Then `writeEndOfWire` could be implemented as
```
    @Override
    public boolean writeEndOfWire(long timeout, TimeUnit timeUnit, long lastPosition) {
        return isEndOfWirePresent(true, timeout, timeUnit, lastPosition) == PresentAfterUpdate;
    }
```

nicktindall: Or even better why not just have `writeEndOfWire` which will write it, and `isEndOfWirePresent` which will ONLY read it?

alamar: Having two of them will be problematic since two subsequent invocations can produce irrelevant results. It will also make it much harder to implement. I'll go with enum.

nicktindall (approved):
Nice

PR #617, opened 2023-02-02
Use expectException rather than serializing an comparing exception keys

nicktindall:
These were from '17 I wonder if that predates `expectException`

### DIFF
--- a/src/test/java/net/openhft/chronicle/wire/UnsupportedChangesTest.java
+++ b/src/test/java/net/openhft/chronicle/wire/UnsupportedChangesTest.java
@@ -19,12 +19,7 @@
 package net.openhft.chronicle.wire;
 
 import net.openhft.chronicle.core.Jvm;
-import net.openhft.chronicle.core.onoes.ExceptionKey;
-import net.openhft.chronicle.core.onoes.LogLevel;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


src/test/java/net/openhft/chronicle/wire/UnsupportedChangesTest.java
```diff
@@ -19,39 +19,27 @@
...
                 "}\n");
         assertEquals("!net.openhft.chronicle.wire.UnsupportedChangesTest$Nested {\n" +
                 "  inner: !!null \"\"\n" +
                 "}\n", nested.toString());
-        ExceptionKey ek = new ExceptionKey(LogLevel.WARN,
-                WireMarshaller.ObjectFieldAccess.class,
-                "Unable to parse field: inner, as a marshallable as it is 128",
-                exceptions.keySet().iterator().next().throwable);
-        assertEquals(Collections.singletonMap(ek, 1), exceptions);
+        expectException("Unable to parse field: inner, as a marshallable as it is 128");
```
nicktindall, line 37: These blew up with the new ExceptionKey, because it had an additional field. Using `expectException` should be less brittle

vach-chronicle (approved):

PR #619, opened 2023-02-05
Don't allow enterHeader to overwrite EOF. Closes #618

### LINKED ISSUE #618: Don't allow enterHeader to overwrite EOF

JerryShea:


src/main/java/net/openhft/chronicle/wire/AbstractWire.java
```diff
@@ -310,6 +310,8 @@ public long enterHeader(long safeLength) {
             if (isNotComplete(header)) {
                 if (header != END_OF_DATA)
                     Jvm.warn().on(getClass(), new Exception("Incomplete header found at pos: " + pos + ": " + Integer.toHexString(header) + ", overwriting"));
+                else
+                    throw new IllegalStateException("You should not be able to write at EOF");
```
JerryShea: enterHeader's behaviour is tested in CQ

nicktindall (approved):

src/main/java/net/openhft/chronicle/wire/AbstractWire.java
nicktindall: Should we have a specific subclass of IllegalStateException? it seems like a pretty specific scenario

JerryShea: Done

PR #644, opened 2023-03-27
Add reproduction and solution for long generated class names. Fixes #643

### LINKED ISSUE #643: Method writers that implement a lot of interfaces sometimes have an impossibly long filename
e.g. `AllOut2OrderRequestListenerHeartbeatSourceHeartbeatDefinitionsListenerMarketDataUpdateListenerMidListenerNewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListenerPeriodicUpdateSourcePositionAdjustmentsListenerPositionServerConfigListenerCalculationLogListenerResetAllPositionsListenerSignalDataListenerPositionListenerStatusListenerConfigOutPricerOutDebuggableApplicationInfoListenerApplicationErrorListenerLadderByQtyListenerLadderByRateListenerHeartbeatListenerHeartbeatErrorListenerHedgerControlListenerExecutionReportListenerEcommTradeListenerEquivalentPositionsListenerDropCopyTradeListenerDropCopyTradeMissingAlertListenerOrderExecutionReportListenereventInterceptingTextMethodWriter`

nicktindall:
This detects when we're about to try and create a filename with > 255 characters ([which is the limit for the linux ext* filesystems](https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits)). And truncates the concatenated list of interfaces, replacing the last ~15 characters with a Base32 encoded (filename safe) 64-bit hash of the original pre-truncation file name.

This should produce deterministic and unique filenames that don't blow the file-system limits.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/VanillaMethodWriterBuilder.java
+++ b/src/main/java/net/openhft/chronicle/wire/VanillaMethodWriterBuilder.java
@@ -21,6 +21,7 @@ import net.openhft.chronicle.bytes.*;
 import net.openhft.chronicle.core.Jvm;
 import net.openhft.chronicle.core.io.Closeable;
 import net.openhft.chronicle.core.util.Builder;
+import net.openhft.chronicle.wire.internal.MethodWriterClassNameGenerator;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -59,6 +60,7 @@ public class VanillaMethodWriterBuilder<T> implements Builder<T>, MethodWriterBu
     private final boolean disableProxyGen = Jvm.getBoolean(DISABLE_WRITER_PROXY_CODEGEN, false);
     private final Set<Class<?>> interfaces = Collections.synchronizedSet(new LinkedHashSet<>());
 
+    private final MethodWriterClassNameGenerator methodWriterClassNameGenerator;
     private final String packageName;
     private ClassLoader classLoader;
     @NotNull
@@ -82,6 +84,7 @@ public class VanillaMethodWriterBuilder<T> implements Builder<T>, MethodWriterBu
         //  Maybe have an option to always use current thread class loader?
         this.classLoader = clsLdr != null ? clsLdr : getClass().getClassLoader();
         this.handlerSupplier = new MethodWriterInvocationHandlerSupplier(handlerSupplier);
+        this.methodWriterClassNameGenerator = new MethodWriterClassNameGenerator();
     }
 
     @NotNull
@@ -149,25 +152,13 @@ public class VanillaMethodWriterBuilder<T> implements Builder<T>, MethodWriterBu
     }
 
     /**
-     * because we cache the classes in {@code classCache}, its very important to come up with a name that is unique for what the class does.
+     * because we cache the classes in {@code classCache}, it's very important to come up with a name that is unique for what the class does.
      *
      * @return the name of the new class
      */
     @NotNull
     private String getClassName() {
-        final StringBuilder sb = new StringBuilder();
-
-        interfaces.forEach(i -> {
-            if (i.getEnclosingClass() != null)
-                sb.append(i.getEnclosingClass().getSimpleName());
-            sb.append(i.getSimpleName());
-        });
-        sb.append(this.genericEvent == null ? "" : this.genericEvent);
-        sb.append(this.metaData ? "MetadataAware" : "");
-        sb.append(updateInterceptor != null ? "Intercepting" : "");
-        sb.append(toFirstCapCase(wireType().toString().replace("_", "")));
-        sb.append("MethodWriter");
-        return sb.toString();
+        return methodWriterClassNameGenerator.getClassName(interfaces, genericEvent, metaData, updateInterceptor != null, wireType());
     }
 
     @NotNull
--- a/src/main/java/net/openhft/chronicle/wire/internal/MethodWriterClassNameGenerator.java
+++ b/src/main/java/net/openhft/chronicle/wire/internal/MethodWriterClassNameGenerator.java
@@ -1,0 +1,49 @@
+package net.openhft.chronicle.wire.internal;
+
+import net.openhft.chronicle.core.Maths;
+import net.openhft.chronicle.wire.Base32LongConverter;
+import net.openhft.chronicle.wire.LongConverter;
+import net.openhft.chronicle.wire.WireType;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Set;
+
+public class MethodWriterClassNameGenerator {
+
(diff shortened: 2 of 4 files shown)


src/main/java/net/openhft/chronicle/wire/VanillaMethodWriterBuilder.java
```diff
@@ -56,6 +57,10 @@
             Observer.class
     );
 
+    private static final LongHashFunction HASH_FUNCTION = LongHashFunction.city_1_1();
```
nicktindall: I didn't put a lot of thought into which hash to use, so happy to change to something else.

src/main/java/net/openhft/chronicle/wire/internal/MethodWriterClassNameGenerator.java
```diff
@@ -0,0 +1,50 @@
...
+            sb.delete(firstIndexTruncated, endOfInterfacesIndex);
+            sb.insert(firstIndexTruncated, Base32LongConverter.INSTANCE.asText(hashOfTruncatedInterfaces));
+        }
+        return sb.toString();
+    }
+
+    @NotNull
+    private String toFirstCapCase(@NotNull String name) {
+        return Character.toUpperCase(name.charAt(0)) + name.substring(1).toLowerCase();
+    }
```
nicktindall, line 48: Extracted out so we can test it properly, also potentially could be configurable if there were ever reason to want different generated names.

src/test/java/net/openhft/chronicle/wire/method/GenerateMethodWriterInheritanceTest.java
```diff
@@ -98,6 +98,18 @@ public void testGenerateForClass() {
...
+    @Test
+    public void testGenerateForLongGeneratedClassName() {
+        final Wire wire = BINARY.apply(Bytes.elasticByteBuffer());
+
+        wire.methodWriter(
+                NewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener1.class,
+                NewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener2.class,
+                NewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener3.class
+        );
+    }
```
nicktindall, line 110: Also integration test

src/main/java/net/openhft/chronicle/wire/VanillaMethodWriterBuilder.java
peter-lawrey: I suggest Maths.hash64(CharSequence) instead

peter-lawrey: It doesn't need to be a complete Hash, you can have the most important details being readable

src/main/java/net/openhft/chronicle/wire/internal/MethodWriterClassNameGenerator.java
peter-lawrey: It's the generated class name you are interested in, that should be testable.

pom.xml
```diff
@@ -94,6 +94,11 @@
             <artifactId>annotations</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>net.openhft</groupId>
+            <artifactId>zero-allocation-hashing</artifactId>
+        </dependency>
+
```
peter-lawrey: Drop this dependency

src/main/java/net/openhft/chronicle/wire/internal/MethodWriterClassNameGenerator.java
nicktindall: It is, that's exactly what MethodWriterClassNameGeneratorTest does. (see [the test](https://github.com/OpenHFT/Chronicle-Wire/pull/644/files#diff-dda75cf859856e30c3a4e3004c0368cb1a23e447196272bfea3e75fdcfda0aa2))

Do you mean you want the package testable as well? I don't think that's as important as it is pretty much static.

src/main/java/net/openhft/chronicle/wire/VanillaMethodWriterBuilder.java
nicktindall: Yes, only the portion which is truncated is replaced by its hash. we keep as much of the interface list as we can and we keep all of `MetadataAware|Intercepting|WireType|etc.`. It's so we can safely truncate without introducing naming collisions for classes that only differ in their truncated portion.

nicktindall: I replaced the zero-allocation-hashing city hash with the one from Maths, but it doesn't support hashing a substring of a StringBuilder so what you get now is the hash of the entire classname pre-truncation instead of the hash of just the bit that was truncated. That's fine though, all the original properties hold.

nicktindall: You can see some examples of what it produces in [the test](https://github.com/OpenHFT/Chronicle-Wire/pull/644/files#diff-dda75cf859856e30c3a4e3004c0368cb1a23e447196272bfea3e75fdcfda0aa2)

src/main/java/net/openhft/chronicle/wire/internal/MethodWriterClassNameGenerator.java
nicktindall: (see [the test](https://github.com/OpenHFT/Chronicle-Wire/pull/644/files#diff-dda75cf859856e30c3a4e3004c0368cb1a23e447196272bfea3e75fdcfda0aa2))

src/test/java/net/openhft/chronicle/wire/internal/MethodWriterClassNameGeneratorTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Test
+    public void testGetClassNameIncludesAllModifiers() {
+        assertEquals("MethodWriterClassNameGeneratorTestJustAnInterfaceFooBarMetadataAwareInterceptingRawMethodWriter",
+                generatedClassName("FooBar", true, true, WireType.RAW, JustAnInterface.class));
+    }
+
+    @Test
+    public void testGetClassNameTruncatesInterfaceNamesWhenMaxFilenameLengthIsExceeded() {
+        assertEquals("MethodWriterClassNameGeneratorTestNewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener1MethodWriterClassNameGeneratorTestNewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGi5JWELPHK3VKNITextMethodWriter",
```
JerryShea, line 50: cool

JerryShea (approved):

PR #648, opened 2023-04-04
Add support for generated yaml tests 

peter-lawrey:


src/main/java/net/openhft/chronicle/wire/AbstractCommonMarshallable.java
```diff
@@ -43,6 +43,8 @@ public String toString() {
         ValidatableUtil.startValidateDisabled();
         try {
             return Marshallable.$toString(this);
+        } catch (Throwable e) {
```
JerryShea, line 46: do we really want to be catching Throwable here, not Exception?

RobAustin (approved):
Would be good to number the test class that it generates, to help easily find them you have to re-run.

src/main/java/net/openhft/chronicle/wire/AbstractCommonMarshallable.java
peter-lawrey: If you are trying to write a warning message and it throws an Error, would you rather see just the error or also see the message it's attempting to write?

PR #650, opened 2023-04-06
Render zero correctly in SymbolsLongConverter. Fixes #649

### LINKED ISSUE #649: There's no way to encode a "zero" value in the SymbolsLongConverters
We take numeric zero to mean empty string which means strings that consist entirely of the "zero" character can't be round-tripped.

e.g. the following both fail
`assertEquals(Base85LongConverter.INSTANCE.asText(Base85LongConverter.INSTANCE.parse("0")), "0");`
`assertEquals(Base85LongConverter.INSTANCE.asText(Base85LongConverter.INSTANCE.parse("00000")), "00000");`

a similar problem exists for values with leading "zero" values e.g.
`assertEquals(Base85LongConverter.INSTANCE.asText(Base85LongConverter.INSTANCE.parse("00012")), "00012");`

Not sure if this is a bug, or just a limitation to be aware of when using them to encode strings. Perhaps it's something we need to be explicit about in our docs?

nicktindall:
See #649 for details

Slight change of behaviour because empty string will not round-trip correctly, but I don't think we should expect it to. I believe parsing an empty string as zero is just an alternative to throwing a `NumberFormatException`. We should still write zeroes as they would be written in a base-whatever encoding.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/converter/SymbolsLongConverter.java
+++ b/src/main/java/net/openhft/chronicle/wire/converter/SymbolsLongConverter.java
@@ -63,6 +63,11 @@ public class SymbolsLongConverter implements LongConverter {
 
     @Override
     public void append(StringBuilder text, long value) {
+        if (value == 0) {
+            text.append(decode[0]);
+            return;
+        }
+
         final int start = text.length();
         if (value < 0) {
             int v = (int) Long.remainderUnsigned(value, factor);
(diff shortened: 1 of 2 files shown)


nicktindall:
Actually looking at the tests that broke it seems important to be able to represent an empty string somehow.

It's confusing because if you treat the encoding as base-XX encoding rendering zero as an empty string would seem like nonsense. But if you think of it primarily as a way of encoding strings it makes perfect sense.

Closing.

PR #654, opened 2023-05-05
Support scanning for expected methods and ignoring the rest

peter-lawrey:


RobAustin (approved):
good but, maybe just missing documentation in .adoc

PR #655, opened 2023-05-10
Verbose types should be carried through to the transport interfaces

RobAustin:
relates to Chronicle-Services/issues/472


src/main/java/net/openhft/chronicle/wire/GenerateMethodWriter.java
```diff
@@ -393,8 +393,8 @@ private Class<?> createClass() {
             imports.append(interfaceMethods);
             imports.append("\n}\n");
 
-            if (DUMP_CODE)
-                System.out.println(imports);
+            //    if (DUMP_CODE)
```
peter-lawrey: Are you sure you want this all the time to sout?

peter-lawrey (approved):

PR #658, opened 2023-05-10
Increased test coverage for YamlWire

peter-lawrey:


alamar:
There seem to be Zing test failures @peter-lawrey

Issue #509, opened 2022-06-17
Update all test to work with YAML wire 


RobAustin:
Update all test to work with YAML wire by using system property wire.testAsYaml, then remove TextWrire, first mark deprecated

alamar:
First phase would be #411 to remove a separate Text wire writer, then we should consider if removing Text wire parser is feasible as well.

RobAustin:
@peter-lawrey wants to remove or rather deprecate TextWire, so can you close the ticket once we get to 100%, best to list the test which still use text wire.

alamar:
I will check which tests only use TEXT wire but not YAML. But thing is, a lot of test cases are actually covered by tests outside Wire (such as Map, Queue tests which take TEXT as input or output) and those only test TEXT wire

So we will know how much stuff is broken only when we switch. I will do that to my TextWire writer unifications and see if anything breaks.

alamar:
TEXT wire still has to be tested since it is actively used. Moreover I'm not sure if we can deprecate TextWire completely since we may be reliant on it for reading existing text documents produced by TextWire which may not be proper YAML.

alamar:
> [ERROR] Tests run: 10228, Failures: 66, Errors: 27, Skipped: 30

alamar:
YAML wire does not support deserialization of true as 1 and false as 0 - do we need it?

TextBinaryWireTest is the relevant test

alamar:
`InvalidYamWithCommonMistakesTest` failures were ignored because they test for non-compliant parsing and the goal of YAML is to be compliant

alamar:
Many tests were added but YAML test coverage and feature coverage is still incomplete.

PR #666, opened 2023-05-16
Update tests to work with YAML wire #509

alamar:
Add test for YAML method writer
Fix compatibility marshallable test
Fix nested maps test
Fix some reordering tests
Fix text binary wire test
Case insensitive field name matching to match TEXT behavior Fix type literal parsing
Fixed handling of sequences
Partially support scenario where not all fields are read Improve test coverage of tricky scenarios


alamar:
https://teamcity.chronicle.software/buildConfiguration/Chronicle_BuildAll_BumpedBranch/805573?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true passes

peter-lawrey (approved):

PR #672, opened 2023-05-24
fixed gateway recursion issue

RobAustin:
image


george-ball (approved):

peter-lawrey (commented):
It's more of a workaround than a fix, but fine for now.

PR #679, opened 2023-06-06
Add a scanning option to VanillaMethodReaderBuilder

peter-lawrey:
This has the changes compared to an older version skipping the intermediate solution of only enabling scanning.
Now scanning is off by default and can be enabled on the builder.


src/main/java/net/openhft/chronicle/wire/YamlWire.java
```diff
@@ -1344,7 +1344,6 @@ public <T, K> WireIn sequence(@NotNull T t, K kls, @NotNull TriConsumer<T, K, Va
             while (true) {
                 switch (yt.current()) {
                     case SEQUENCE_ENTRY:
-                        yt.next(Integer.MIN_VALUE);
```
JerryShea: what are these changes?

peter-lawrey: This change and another was not needed for this feature

JerryShea (approved):

PR #681, opened 2023-06-07
POC: Scoped thread local (moved to core on approval)

### LINKED ISSUE #683: deepCopy fails when copying DTOs with @AsMarshallable dynamic enum fields
It results in an invalid re-use pattern of the shared `Wires.acquireBinaryWire()`

nicktindall:
This fixes #683

It would probably live in core

This eliminates a category of bugs, might need some fine tuning for garbage

Turn off whitespace to review

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireInternal.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireInternal.java
@@ -57,7 +57,7 @@ public enum WireInternal {
     static final ThreadLocal<WeakReference<Bytes<?>>> INTERNAL_BYTES_TL = new ThreadLocal<>();
     static final ScopedThreadLocal<Wire> BINARY_WIRE_SCOPED_TL = new ScopedThreadLocal<>(
             () -> new BinaryWire(Wires.unmonitoredDirectBytes())
-                    .setOverrideSelfDescribing(true), Wire::clear, 4);
+                    .setOverrideSelfDescribing(true), Wire::clear, 4, true);
 
     static final StackTraceElement[] NO_STE = {};
     static final Set<Class> INTERNABLE = new HashSet<>(Arrays.asList(
--- a/src/main/java/net/openhft/chronicle/wire/WireInternal.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireInternal.java
@@ -29,6 +29,7 @@ import net.openhft.chronicle.core.pool.StringInterner;
 import net.openhft.chronicle.core.threads.ThreadLocalHelper;
 import net.openhft.chronicle.core.util.*;
 import net.openhft.chronicle.wire.internal.FromStringInterner;
+import net.openhft.chronicle.wire.scoped.ScopedThreadLocal;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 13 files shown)


src/main/java/net/openhft/chronicle/wire/scoped/ScopedThreadLocal.java
```diff
@@ -0,0 +1,62 @@
+package net.openhft.chronicle.wire.scoped;
+
+import net.openhft.chronicle.core.Jvm;
+
+import java.lang.reflect.Array;
+import java.util.function.Supplier;
+
+public class ScopedThreadLocal<T> {
```
alamar, line 28: Please add some javadoc - it's hard to understand what's going on. Perhaps it makes sense to ask @peter-k-lawrey and then put it into Chronicle-Core directly

src/main/java/net/openhft/chronicle/wire/WireInternal.java
```diff
@@ -54,6 +55,9 @@ public enum WireInternal {
     static final ThreadLocal<WeakReference<Bytes<?>>> BYTES_F2S_TL = new ThreadLocal<>();
     static final ThreadLocal<WeakReference<Wire>> BINARY_WIRE_TL = new ThreadLocal<>();
     static final ThreadLocal<WeakReference<Bytes<?>>> INTERNAL_BYTES_TL = new ThreadLocal<>();
+    static final ScopedThreadLocal<Wire> BINARY_WIRE_SCOPED_TL = new ScopedThreadLocal<>(
+            () -> new BinaryWire(Wires.unmonitoredDirectBytes())
+                    .setOverrideSelfDescribing(true), Wire::clear, 4, true);
```
JerryShea: when we productionise, the max depth should be parameterisable

JerryShea (approved):
Love this POC. Fully support the concept

src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -399,6 +400,10 @@ public static Wire acquireBinaryWire() {
         return wire;
     }
 
+    public static ScopedResource<Wire> acquireBinaryWireScoped() {
+        return WireInternal.BINARY_WIRE_SCOPED_TL.get();
+    }
+
     @NotNull
     public static Bytes<?> acquireAnotherBytes() {
         if (Jvm.isDebug())
```
JerryShea, line 409: applying this new approach uniformly will enable us to remove this different behaviour when running in debugger. @peter-lawrey I know you did this deliberately - there are comments like "when in debug, the output becomes confused if you reuse the buffer" but it seems sub-optimal to change behaviour of code when running in the debugger, and you would hope that this new approach would ensure that no buffers will get erroneously reused

src/main/java/net/openhft/chronicle/wire/scoped/ScopedThreadLocal.java
```diff
@@ -0,0 +1,157 @@
...
+     * @return the {@link ScopedResource}, to be closed once it is finished being used
+     */
+    public ScopedResource<T> get() {
+        final SimpleStack scopedThreadLocalResources = instancesTL.get();
+        AbstractScopedResource<T> instance;
+        if (scopedThreadLocalResources.isEmpty()) {
+            instance = createNewResource();
+        } else {
+            instance = scopedThreadLocalResources.pop();
+        }
```
nicktindall, line 83: Note here that the instance created/acquired is removed from the `SimpleStack`, so if the `get()`-ter doesn't return the instance (by calling `ScopedResource#close()`), it'll end up being GC'd. If this is approved I might add a `Cleaner` that confirms the resource was discarded correctly and warns otherwise.

nicktindall: The alternative would be keeping a reference to it when it's checked out, but that could lead to resource leaks. I think the instance being GC'd rather than being returned to the pool is probably preferable.

src/main/java/net/openhft/chronicle/wire/Wires.java
peter-lawrey: It's sub-optimal when running debugging but more optimal when not running debugging. This change would allow us to remove such checks. The small overhead isn't a problem if rarely used.

peter-lawrey (approved):

nicktindall:
This has been implemented here now https://github.com/OpenHFT/Chronicle-Core/pull/576

PR #686, opened 2023-06-09
Temp/remove reset

JerryShea:
WIP don't merge

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/AbstractMarshallableCfg.java
+++ b/src/main/java/net/openhft/chronicle/wire/AbstractMarshallableCfg.java
@@ -27,7 +27,7 @@ public abstract class AbstractMarshallableCfg extends SelfDescribingMarshallable
     @Override
     public void readMarshallable(@NotNull WireIn wire) throws IORuntimeException, InvalidMarshallableException {
         WireMarshaller wm = WireMarshaller.WIRE_MARSHALLER_CL.get(this.getClass());
-        wm.readMarshallable(this, wire, wm.defaultValue(), false);
+        wm.readMarshallable(this, wire, wm.defaultValue(), true);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


src/main/java/net/openhft/chronicle/wire/AbstractMarshallableCfg.java
```diff
@@ -27,7 +27,7 @@ public abstract class AbstractMarshallableCfg extends SelfDescribingMarshallable
     @Override
     public void readMarshallable(@NotNull WireIn wire) throws IORuntimeException, InvalidMarshallableException {
         WireMarshaller wm = WireMarshaller.WIRE_MARSHALLER_CL.get(this.getClass());
-        wm.readMarshallable(this, wire, wm.defaultValue(), false);
+        wm.readMarshallable(this, wire, wm.defaultValue(), true);
```
nicktindall, line 30: Might be worth doing a build-all branch to see what effect this has...

JerryShea: already kicked off :)

src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -603,9 +603,6 @@ public static <E> E object0(ValueIn in, @Nullable E using, @Nullable Class clazz
     }
 
     public static <E> E object1(ValueIn in, @Nullable E using, @Nullable Class clazz, boolean bestEffort) throws InvalidMarshallableException {
-        if (using instanceof AbstractMarshallableCfg)
```
JerryShea, line 606: undid this but not the associated test. I believe that the change to AbstractMarshallableCfg is more intentional

peter-lawrey (approved):

PR #689, opened 2023-06-26
Add an option of using the wall clock for timings.

peter-lawrey:
Add recordHistory() option to ChronicleChannels

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
+++ b/src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
@@ -181,7 +181,6 @@ public class VanillaMessageHistory extends SelfDescribingMarshallable implements
             wire.read("sources").sequence(this, VanillaMessageHistory::acceptSourcesRead);
             timings = 0;
             wire.read("timings").sequence(this, VanillaMessageHistory::acceptTimingsRead);
-            wallClock = wire.read("wallClock").bool();
         }
         if (addSourceDetails) {
             @Nullable Object o = wire.parent();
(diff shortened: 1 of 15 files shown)


RobAustin (approved):

src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
```diff
@@ -164,6 +181,7 @@ public void readMarshallable(@NotNull WireIn wire) throws IORuntimeException, In
             wire.read("sources").sequence(this, VanillaMessageHistory::acceptSourcesRead);
             timings = 0;
             wire.read("timings").sequence(this, VanillaMessageHistory::acceptTimingsRead);
+            wallClock = wire.read("wallClock").bool();
```
JerryShea: can't we just configure globally for the system rather than change wire format?

src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
```diff
@@ -205,6 +221,8 @@ private void readMarshallable0(@NotNull BytesIn<?> bytes) {
         timings = bytes.readUnsignedByte();
         for (int i = 0; i < timings; i++)
             timingsArray[i] = bytes.readLong();
+        if (bytes.readRemaining() > 0)
```
JerryShea, line 224: please undo these 2 lines

src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
```diff
@@ -53,6 +58,7 @@ public class VanillaMessageHistory extends SelfDescribingMarshallable implements
     private int sources;
     private int timings;
     private boolean addSourceDetails = false;
+    private boolean wallClock = HISTORY_WALL_CLOCK;
```
JerryShea, line 61: should be final?

JerryShea: Do we actually need `wallClock` field - maybe all we need is a `LongSupplier`  `wallClock ? CLOCK.currentTimeNanos() : System.nanoTime();` then we won't need the conditional in the `nanoTime` method

JerryShea:
replaced by #731

PR #696, opened 2023-07-05
BufferedChronicleChannel to throwExceptionIfClosed when setting eventPoller

RobAustin:
Before this change, it was possible to set the eventPoller after the BufferedChronicleChannel was closed, so the eventPoller was not getting cleaned up.


peter-lawrey (approved):
Good catch.

PR #697, opened 2023-07-11
added support for multiple connections to a client channel

RobAustin:
it's now possible to add more than one host port to a client channel 

```
channelCfg.addHostnamePort("localhost", 8092);
```

for example:

```
/**
     * This test verifies the behavior of the system when the primary server is unavailable. It sets up a server and a client
     * that attempt to connect to two different servers. The first server (primary) is intentionally configured to be unavailable.
     * After the client fails to connect to the primary server, it attempts to connect to the second server (secondary).
     * The secondary server is the host port configured for the gateway to run on.
     */
    @Test
    public void testEchoHandlerOnSecondaryConnection() {
        try (ChronicleContext context = ChronicleContext.newContext("tcp://:0?sessionName=testId")) {
            context.startNewGateway();

            final ChronicleChannelCfg channelCfg = new ChronicleChannelCfg().initiator(true).buffered(true);

            // this will be invalid
            channelCfg.addHostnamePort("localhost", 8092);

            // this should be valid
            channelCfg.addHostnamePort("localhost", context.port());

            try (ChronicleChannel channel = ChronicleChannel.newChannel(context.socketRegistry(), channelCfg,
                    new EchoHandler())) {
                channel.methodWriter(SayMsg.class).say(EXPECTED);

                final StringBuilder eventType = new StringBuilder();
                String actual = channel.readOne(eventType, String.class);
                assertEquals(EXPECTED, actual);
            }
        }
    }
```


src/main/java/net/openhft/chronicle/wire/channel/ChronicleContext.java
```diff
@@ -137,6 +142,27 @@ public static URL urlFor(String spec) throws IORuntimeException {
     }
 
     /**
+     * Obtain a URL from a String, loading any custom Handler as needed.
+     *
+     * @param spec to use for the URL
+     * @return the URL
+     * @throws IORuntimeException if malformed
+     */
+    public static List<URL> urlsFor(String spec) throws IORuntimeException {
```
peter-lawrey: This needs revising. I would drop it from this PR for now.

peter-lawrey (changes requested):
Can you remove the support for multiple URL in the same string? 
The addHostnamePort method is good and enough for now.

peter-lawrey (approved):

PR #706, opened 2023-08-07
Remove padding after the end of document ... on a close.

peter-lawrey:
This way isEmpty() won't return true when there is nothing to read.


peter-lawrey-admin (approved):
Needs to be fixed urgently

PR #707, opened 2023-08-07
Sonar/add an assert

peter-lawrey:


src/test/java/net/openhft/chronicle/wire/method/GenerateMethodWriterInheritanceTest.java
```diff
@@ -102,14 +103,16 @@ public void testGenerateForClass() {
...
 
-        wire.methodWriter(
+        Object writer = wire.methodWriter(
                 NewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener1.class,
                 NewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener2.class,
                 NewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener3.class
         );
+        assertInstanceOf(NewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener1.class, writer);
+        assertInstanceOf(NewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener2.class, writer);
+        assertInstanceOf(NewOrderSingleListenerOmsHedgerTradeListenerOpenOrdersListenerPaidGivenTickListener3.class, writer);
```
nicktindall, line 113: I pushed a change to use `assertInstanceOf(x, y)` instead of `assertTrue(x instanceof y)` as it allows JUnit to give a better error message when it fails.

nicktindall (approved):
Please squash before rebasing if you're happy with the changes, also if sonar complains about something and we know it's a false positive, we can use `SuppressWarnings` e.g. for this example `@SuppressWarnings("java:S2699")`.

That test would fail before the fix I introduced because the compilation of the class failed and spewed out warnings which, because the test extends `WireTestCommon`, break the build.

We fail this check a lot because often passing means not producing any warnings. Certainly asserting the thing you're testing is the best practice and most explicit approach, but in this case that's not what we're asserting.

peter-lawrey:
@nicktindall I did a `git rebase -i origin/develop` with a `git push --force-with-lease`, but I didn't see much change.
In any case, github defaults to "Squash and merge"

PR #719, opened 2023-08-25
Inline deprecated Maths.hash32 method

peter-lawrey:


RobAustin (approved):
approved assuming that the entries[h2] are not persisted, as it looks like the new code changes the hash function.

PR #721, opened 2023-08-30
throw an IAE on a missing class alias

peter-lawrey:


JerryShea (approved):
Love it

src/main/java/net/openhft/chronicle/wire/TextWire.java
```diff
@@ -66,6 +66,8 @@ public class TextWire extends YamlWireOut<TextWire> {
     static final Supplier<StopCharsTester> STRICT_END_OF_TEXT_ESCAPING = TextStopCharsTesters.STRICT_END_OF_TEXT::escaping;
     static final Supplier<StopCharsTester> END_EVENT_NAME_ESCAPING = TextStopCharsTesters.END_EVENT_NAME::escaping;
     static final Bytes<?> META_DATA = Bytes.from("!!meta-data");
+    @Deprecated(/* for removal in x.26, make default true in x.25 */)
+    static final boolean IAE_ON_CNF = Jvm.getBoolean("illegal.argument.for.missing.class.alias", false);
```
JerryShea, line 70: @chroniclekevinpowe given that this will not change behaviour unless someone turns on the system property, I think that this is a candidate for backporting to 23

peter-lawrey: It should be possible to backport even breaking changes with this strategy of having it disabled by default.

PR #727, opened 2023-09-19
Deprecate multitude of singleton thread-local resource, use ScopedThreadLocal instead.

### LINKED ISSUE #683: deepCopy fails when copying DTOs with @AsMarshallable dynamic enum fields

nicktindall:
Depends on OpenHFT/Chronicle-Core#576

Fixes #683

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
@@ -21,6 +21,7 @@ import net.openhft.chronicle.bytes.Bytes;
 import net.openhft.chronicle.bytes.HexDumpBytesDescription;
 import net.openhft.chronicle.core.*;
 import net.openhft.chronicle.core.io.*;
+import net.openhft.chronicle.core.scoped.ScopedResource;
 import net.openhft.chronicle.core.pool.StringBuilderPool;
 import net.openhft.chronicle.core.util.ClassNotFoundRuntimeException;
 import net.openhft.chronicle.core.util.ObjectUtils;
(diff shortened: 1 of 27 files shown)


src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -1211,12 +1216,14 @@ protected void setValue(Object o, @NotNull ValueIn read, boolean overwrite) {
...
-            byte[] decode = Base64.getDecoder().decode(s);
-            bytes.clear();
-            bytes.write(decode);
+            try (ScopedResource<StringBuilder> stlSb = Wires.acquireStringBuilderScoped()) {
+                @NotNull StringBuilder sb0 = stlSb.get();
+                read.text(sb0);
+                String s = WireInternal.INTERNER.intern(sb0);
+                byte[] decode = Base64.getDecoder().decode(s);
+                bytes.clear();
+                bytes.write(decode);
```
nicktindall, line 1246: I wonder how often interning works with arbitrary base64 strings, do we not have a utility to decode base64 directly from a StringBuilder?

src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -1055,15 +1122,17 @@ enum SerializeBytes implements Function<Class, SerializationStrategy> {
...
+                @NotNull StringBuilder sb0 = stlSb.get();
+                in.text(sb0);
+                String s = WireInternal.INTERNER.intern(sb0);
+                byte[] decode = Base64.getDecoder().decode(s);
+                if (o == null)
+                    return Bytes.wrapForRead(decode);
+                o.clear();
+                o.write(decode);
+                return o;
+            }
```
nicktindall, line 1143: again it'd be great to be able to base64 decode directly from a string builder

JerryShea (approved):

PR #729, opened 2023-09-21
DRAFT: Replace literal 4 with header constant or Integer.BYTES

tgd:
Nick, this is a draft PR to begin replacing literal 4's with either the header constant or Integer.BYTES. In some cases it is a little unclear whether the header constant should be used. Let me know your thoughts.


nicktindall:
I'm actually not that confident in reviewing this. I reckon we should only make this change when we can pad out the header (i.e. change `SBP_HEADER_SIZE` to 100 and verify everything still works). Only then can we be sure that we've used 4s where 4s are required and header_size where it is required.

PR #731, opened 2023-10-06
Add an option of using the wall clock for timings.

JerryShea:
Add recordHistory() option to ChronicleChannels


JerryShea:
This is a rebased version of #689 with code review comments applied. @peter-lawrey are you OK to approve?

peter-lawrey (approved):

PR #734, opened 2023-10-13
parameter to allow history to be written as method id

JerryShea:


src/test/java/net/openhft/chronicle/wire/MessageHistoryTest.java
```diff
@@ -43,10 +43,7 @@ public void checkHistoryGetClear() {
     @Test
     public void checkDeepCopy() {
         VanillaMessageHistory history = new VanillaMessageHistory();
-        history.addSource(1, 0xff);
```
JerryShea, line 46: rationalise and tidy tests

src/main/java/net/openhft/chronicle/wire/MessageHistory.java
```diff
@@ -44,9 +45,10 @@ static void set(MessageHistory md) {
     @UsedViaReflection
     static void writeHistory(DocumentContext dc) {
         Wire wire = dc.wire();
-        if (((WriteDocumentContext) dc).isEmpty()) // only add to the start of a message. i.e. for chained calls.
-            wire.writeEventName(MethodReader.HISTORY)
-                    .marshallable(get());
+        if (((WriteDocumentContext) dc).isEmpty()) { // only add to the start of a message. i.e. for chained calls.
+            final ValueOut valueOut = HISTORY_METHOD_ID ? wire.writeEventId(MethodReader.MESSAGE_HISTORY_METHOD_ID) : wire.writeEventName(MethodReader.HISTORY);
+            valueOut.marshallable(get());
+        }
```
nicktindall, line 48: If we delegated to the MessageHistory instance to do the write, it could be configured there e.g. `writeHistory` becomes

```
if (((WriteDocumentContext) dc).isEmpty()) {
    MessageHistory tlInst = get();
    tlInst.doWriteHistory(dc);
}
```
And `VanillaMessageHistory` implements

```
public void doWriteHistory(DocumentContext dc) {
    final ValueOut valueOut = this.historyMethodId ? wire.writeEventId(MethodReader.MESSAGE_HISTORY_METHOD_ID) : wire.writeEventName(MethodReader.HISTORY);
            valueOut.marshallable(this);    
}
```

That would make it testable, and remove the need for the static on the interface. Also you see there's similar config already in VanillaMessageHistory e.g. `USE_BYTES_MARSHALLABLE` and `HISTORY_WALL_CLOCK`.

nicktindall: I guess the thing is, if we envisaged that there might be numerous types of `MessageHistory` in use in a process at any time, configuring it here means they'd all write the entry with the same key. So perhaps there's an argument for putting that logic outside the MessageHistory implementation.
Feels a bit smelly though having it configured on the interface like this. And it's not easily testable. Non-testable code is always hard to test.

peter-lawrey (approved):

src/main/java/net/openhft/chronicle/wire/MessageHistory.java
JerryShea: Yep good call @nicktindall - keeping me honest!

PR #736, opened 2023-10-15
fix benchmarks compilation

JerryShea:


marshallingperf/pom.xml
```diff
@@ -183,6 +183,17 @@
...
                             </execution>
+                            <execution>
+                                <id>JSONWireMarshallableMain</id>
+                                <phase>test</phase>
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <executable>${java.home}/bin/java</executable>
+                                    <commandlineArgs>${jvm.requiredArgs} -classpath %classpath run.chronicle.wire.perf.JSONWireMarshallableMain</commandlineArgs>
```
JerryShea, line 194: add JSONWire test to TeamCity charts

peter-lawrey (approved):
Make sure that if affinity is used,. threads(1) or they all have affinity for the same CPU

PR #740, opened 2023-10-18
Use recursive reset for non-primitive fields inside Wires.reset().

### LINKED ISSUE #732: reset() should handle mutable fields, especially collections and marshallables
```
public static void main(String[] args) {
    ExecutionReportDto or1 = new ExecutionReportDto();
    ExecutionReportDto or2 = new ExecutionReportDto();
    Jvm.warn().on(Main.class, String.format("Before reset: %s == %s is %s", or1.clientIdentifier(), or2.clientIdentifier(), or1.clientIdentifier() == or2.clientIdentifier()));

    or1.reset();
    or1.clientIdentifier().clientId("foo");
    Jvm.warn().on(Main.class, String.format("After reset first: %s == %s is %s", or1.clientIdentifier(), or2.clientIdentifier(), or1.clientIdentifier() == or2.clientIdentifier()));

    or2.reset();
    or2.clientIdentifier().requestId("bar");
    Jvm.warn().on(Main.class, String.format("After reset second %s == %s is %s", or1.clientIdentifier(), or2.clientIdentifier(), or1.clientIdentifier() == or2.clientIdentifier()));
    Jvm.park();
}
```

leads to the output:

```
Before reset: ClientIdentifierDto{clientId='null', requestId='null'} == ClientIdentifierDto{clientId='null', requestId='null'} is false
After reset first: ClientIdentifierDto{clientId='foo', requestId='null'} == ClientIdentifierDto{clientId='null', requestId='null'} is false
After reset second ClientIdentifierDto{clientId='foo', requestId='bar'} == ClientIdentifierDto{clientId='foo', requestId='bar'} is true 
```

So `reset()` doesn't really clear the fields of internal objects. It will only work for immutable objects, but not for `SelfDescribingMarshallable` fields. State from one execution report is getting into the second execution report, and even if there was just one instance of `ExecutionReport` it would still fail to reset.

How I imagine this should work:
- For primitive and immutable type fields, replace the value/reference to cached initial value.
- For collections, issue `clear()`. Alternatively, do a `deepCopy(cachedInitialValue, value)`.
- For `SelfDescribingMarshallable`s (and perhaps other kinds of marshallables), issue `reset()`. Alternatively, do a deep copy of cached initial value

But the approach is to be discussed.

yevgenp:
Fixes #732

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
@@ -488,7 +488,13 @@ public class WireMarshaller<T> {
     public void reset(T o) {
         try {
             for (FieldAccess field : fields) {
-                field.copy(defaultValue, o);
+                if (field.isResettable()) {
+                    Object value = field.field.get(o);
+                    if (value != null)
+                        Wires.reset(value);
+                } else {
+                    field.copy(defaultValue, o);
+                }
             }
         } catch (IllegalAccessException e) {
             // should never happen as the types should match.
@@ -980,6 +986,11 @@ public class WireMarshaller<T> {
                 return false;
             }
         }
+
+        protected boolean isResettable() {
+            return false;
+        }
+
     }
 
     static class IntValueAccess extends FieldAccess {
@@ -1105,6 +1116,12 @@ public class WireMarshaller<T> {
         public void getAsBytes(Object o, @NotNull Bytes<?> bytes) throws IllegalAccessException {
             bytes.writeUtf8(String.valueOf(field.get(o)));
         }
+
+        @Override
+        protected boolean isResettable() {
+            return true;
+        }
+
     }
 
     static class StringFieldAccess extends FieldAccess {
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -488,14 +488,26 @@ public T defaultValue() {
...
+                } else
+                    field.copy(defaultValue, o);
             }
         } catch (IllegalAccessException e) {
             // should never happen as the types should match.
             throw new AssertionError(e);
         }
     }
 
+    private static boolean canBeReset(FieldAccess field) {
```
peter-lawrey: This feels like it should be a method in FieldAccess so it can be optimised.
`field.resetWith(defaultValue, o)`

yevgenp: Refactored

peter-lawrey (approved):

maxim-ponomarev (approved):

PR #743, opened 2023-10-18
Reproduce segfault from EFX build

nicktindall:
This is a minimal reproduction of the issue with the EFX build

I don't know what the intention was here, but it seems wrong to traverse into object fields and attempt to recursively reset their fields?

i.e. if I have a DTO with a LocalDate field I probably want that set to null when I reset, not set to a husk of a LocalDate instance (which is immutable and shouldn't be modified) with year=0, month=0 and day=0

Perhaps `ObjectFieldAccess.isResettable()` should only return true for fields that implement `Marshallable`?

It feels like this is way too broad

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
@@ -1058,7 +1058,7 @@ public class WireMarshaller<T> {
             super(field, isLeaf);
             asMarshallable = Jvm.findAnnotation(field, AsMarshallable.class);
             type = field.getType();
-            resettable = !DynamicEnum.class.isAssignableFrom(type);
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -1058,7 +1058,7 @@ static class ObjectFieldAccess extends FieldAccess {
             super(field, isLeaf);
             asMarshallable = Jvm.findAnnotation(field, AsMarshallable.class);
             type = field.getType();
-            resettable = !DynamicEnum.class.isAssignableFrom(type);
+            resettable = !DynamicEnum.class.isAssignableFrom(type) && Marshallable.class.isAssignableFrom(type);
```
nicktindall, line 1061: temporary fix to unblock build-all

JerryShea (approved):

PR #747, opened 2023-10-19
Undo regression from scoped-thread-locals change

nicktindall:
I though this was equivalent to existing behaviour but clearly it's not. Reverted to using a dedicated pool until the method is deprecated.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/Wires.java
+++ b/src/main/java/net/openhft/chronicle/wire/Wires.java
@@ -100,7 +100,9 @@ public enum Wires {
             throw new IllegalStateException(e);
         }
     });
-    static final ScopedResourcePool<StringBuilder> SBP = StringBuilderPool.createThreadLocal();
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -323,15 +325,12 @@ public static StringBuilder acquireStringBuilder() {
         if (Jvm.isDebug()) {
             return new StringBuilder();
         } else {
-            // Dangerous, but equivalent to existing behaviour
-            try (final ScopedResource<StringBuilder> scopedResource = SBP.get()) {
-                return scopedResource.get();
-            }
+            return SBP.acquireStringBuilder();
```
nicktindall, line 328: It's actually not equivalent because the single SB pool is used more broadly than it was. We will retain a specific pool for this use case until the method is deprecated.

JerryShea (approved):

PR #748, opened 2023-10-19
Modified check for "class.not.found.for.missing.class.alias"

chroniclekevinpowe:
Check for the flag is now done immediately on ClassNotFoundException occurring in TextWire#typePrefixOrObject.

(due to the tClass parameter being set to the expected type of the field being set, even on a ClassNotFoundException)

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/TextWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/TextWire.java
@@ -46,7 +46,8 @@ import java.util.regex.Pattern;
 import static java.nio.charset.StandardCharsets.ISO_8859_1;
 import static net.openhft.chronicle.bytes.NativeBytes.nativeBytes;
 import static net.openhft.chronicle.wire.TextStopCharTesters.END_OF_TYPE;
-import static net.openhft.chronicle.wire.Wires.THROW_CNF;
+import static net.openhft.chronicle.wire.Wires.GENERATE_TUPLES;
+import static net.openhft.chronicle.wire.Wires.THROW_CNFRE;
 
 /**
  * YAML Based wire format
@@ -1857,7 +1858,7 @@ public class TextWire extends YamlWireOut<TextWire> {
                     return classLookup().forName(stringBuilder);
                 } catch (ClassNotFoundRuntimeException e) {
                     String message = "Unable to find " + stringBuilder + " " + e.getCause();
-                    if (THROW_CNF)
+                    if (THROW_CNFRE)
                         throw new IllegalArgumentException(message);
                     Jvm.warn().on(getClass(), message);
                     return null;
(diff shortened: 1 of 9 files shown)


src/main/java/net/openhft/chronicle/wire/TextWire.java
```diff
@@ -1881,13 +1881,16 @@ public Object typePrefixOrObject(Class tClass) {
                 try {
                     return classLookup().forName(stringBuilder);
                 } catch (ClassNotFoundRuntimeException e) {
+                    // when populating an object from config, tClass will be set to the expected type for the object's field
+                    String message = "Unable to load " + stringBuilder + ", is a class alias missing.";
```
JerryShea: we don't want to do this if GENERATE_TUPLES

src/test/java/net/openhft/chronicle/wire/marshallable/CNFREOnMissingClassTest.java
```diff
@@ -35,4 +35,23 @@ public void throwClassNotFoundRuntimeExceptionOnMissingClassForField() {
...
+        private TwoFields bothFields;
+        private String name;
+    }
+
+    /**
+     * Failing to load a class for a field with a type of java.lang.Object causes the correct behaviour but in
+     * an unexpected code path (the check for a classloader at TextWire#typeOrPrefixObject - line 1913
+     */
+    @Test(expected = ClassNotFoundRuntimeException.class)
+    public void throwClassNotFoundRuntimeExceptionOnMissingClassForFieldNotObject() {
```
JerryShea, line 48: @peter-lawrey this is the failing test, and this is the customer use case

peter-lawrey (approved):

src/test/java/net/openhft/chronicle/wire/marshallable/CNFREOnMissingClassTest.java
peter-lawrey: Fixed

src/test/java/net/openhft/chronicle/wire/WiresTest.java
```diff
@@ -154,7 +156,76 @@ public void unknownType2() {
...
+
+        assertEquals("!FourValues {\n" +
+                "  string: Hello World,\n" +
+                "  num: !int 1234,\n" +
+                "  big: 0.128,\n" +
+                "  also: extra\n" +
+                "}\n", tv.toString());
+    }
+
+    @Test(expected = ClassNotFoundRuntimeException.class)
```
JerryShea, line 194: Suggest that for new code we use `assertThrows` as this gives finer grained control over where the exception thrown from. Just a suggestion - I don't want to hold things up.

JerryShea (approved):

PR #750, opened 2023-10-19
Use recursive reset for non-primitive fields inside Wires.reset().

yevgenp:


JerryShea:
Closing as this was subsequently fixed another way

PR #752, opened 2023-10-19
Only deep reset if default object is of the same type as current and is not null

alamar:


src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -486,13 +485,16 @@ public T defaultValue() {
     public void reset(T o) {
         try {
             for (FieldAccess field : fields) {
```
JerryShea, line 487: why don't we see if o is Resettable and then call o.reset() ?

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -486,13 +485,16 @@ public T defaultValue() {
...
-                if (field.isResettable()) {
-                    Object value = field.field.get(o);
-                    if (value != null)
+                Class<?> resettableClass = field.resettableClass(defaultValue); // TODO FieldAccess should hold its defaultValue
+                if (resettableClass != null) {
+                    Object value = field.field.get(o); // TODO Replace with unsafe access
+                    if (value != null && resettableClass == value.getClass()) {
```
JerryShea: if value instanceof Resettable call reset() on it?

alamar: Makes sense changing the check to `Resettable`.

alamar:
@JerryShea @nicktindall @yevgenp What do you think of this approach?

Notably, it does not seem that we can practically deep reset collections, as .iterator() is a garbage producing operation.

JerryShea:
> @JerryShea @nicktindall @yevgenp What do you think of this approach?
> 
> Notably, it does not seem that we can practically deep reset collections, as .iterator() is a garbage producing operation.

I don't think we should deep reset.

The user can override `reset()` for a DTO containing child DTOs, and [current behaviour](https://github.com/OpenHFT/Chronicle-Wire#a-note-on-wiresreset) is documented, even though it is confusing

nicktindall:
> @JerryShea @nicktindall @yevgenp What do you think of this approach?
> 
> Notably, it does not seem that we can practically deep reset collections, as .iterator() is a garbage producing operation.

I haven't really given it the review it deserves (it's hard to know what's going on without really checking the code out and  having a look). But I do think the described behaviour is problematic and we should strive to do better there.

One thing this kind of change would require IMO is a fairly extensive barrage of tests, and probably a description of how it works (to replace the paragraph explaining the buggy behaviour), it's a complex problem to solve and I'd be more comfortable if we agreed on some behaviour and then wrote up tests to confirm that the behaviour was implemented faithfully.

src/test/java/net/openhft/chronicle/wire/WireResetTest.java
```diff
@@ -72,7 +70,7 @@ public void testDeepReset() {
         assertSame(identifier1, event1.identifier);
         assertNull(event1.identifier.id);
         assertTrue(event1.identifier.permissions.isEmpty());
-        assertNull(event1.identifier.parent.id);
+        assertNull(event1.identifier.parent);
```
yevgenp, line 73: That's not a **deep** reset anymore. And is causing a GC.

alamar: @yevgenp We need to return objects to their pristine (post-constructor) state. If post-constructed field value is null then we need to restore it to null also.
User can have different lifecycle for this field as to avoid GC.

alamar: Note that it's not the only possibility, it's that we struggle to get team approval even from this toned down change. I encourage you to join the discussion on dev@.

alamar: (note that in this case, resetting `parent` when child object is reset would definitely be a bug. Parent usually has multiple children and the subtree should not be all reset when only one child's `reset()` is called)

alamar:
https://teamcity.chronicle.software/buildConfiguration/Chronicle_BuildAll_BumpedBranch/876320?buildTab=overview&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true passed

README.adoc
```diff
@@ -739,7 +739,7 @@ However, more complicated objects with object instance fields may experience und
 In order to `reset` a marshallable object, the process is as follows:
 
 1. create a new instance of the object to be reset
-2. copy all fields from the new instance to the existing instance
+2. copy all fields by reference from the new instance to the existing instance (but see below)
```
maxim-ponomarev: Maybe we could provide a paragraph about "exception" right here (instead of making a footnote)?

README.adoc
```diff
@@ -739,7 +739,7 @@ However, more complicated objects with object instance fields may experience und
 In order to `reset` a marshallable object, the process is as follows:
 
 1. create a new instance of the object to be reset
```
maxim-ponomarev: It sounds like we _**always**_ create a new instance of the to be reset, which is

1. Not true (AFAIK we use cached value for that)?
2. Doesn't really align with our general "garbage free" approach.

maxim-ponomarev (approved):

README.adoc
```diff
@@ -781,6 +781,8 @@ true
 
 showing that the field **b** of each container object is now referencing the same `ByteBuffer` instance.
 
+An exception to this is when the field value implements `Resettable` interface, when the default value of the field is not null, and the current field value is of exactly the same class as the default value. In this case, `value.reset()` will be called instead of overwriting it with default value by reference.
+
 In order to work around this, if necessary, the marshallable class can override `Marshallable.reset`:
```
maxim-ponomarev, line 790: Can we change an example below, do demonstrate, that now it's possible to change field class to `Resettable`, in order to:

1. Avoid having an undesired side-effect on reset.
2. Support a garbage-free behaviour out-of-the-box.

README.adoc
maxim-ponomarev: We can also mention, that resetting Resettable is a new behaviour, introduced in `*.25`.

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -1113,15 +1153,27 @@ protected void setValue(Object o, @NotNull ValueIn read, boolean overwrite) thro
         }
 
         @Override
-        public void getAsBytes(Object o, @NotNull Bytes<?> bytes) throws IllegalAccessException {
-            bytes.writeUtf8(String.valueOf(field.get(o)));
+        protected void setDefaultValue(Object ignored, Object o) {
+            if (defaultValue != null && defaultValue instanceof Resettable && !(defaultValue instanceof DynamicEnum)) {
```
nicktindall: Should we have a specific `ResettableObjectFieldAccess`, to de-clutter the `ObjectFieldAccess`? instanceof strikes me as a smell here.

E.g. you could just have `ResettableObjectFieldAccess` which decorates `ObjectFieldAccess` to add the additional behaviour and is only used when `defaultValue` is not null, `Resettable` and not a `DynamicEnum`

I believe @JerryShea went on a rampage once because using instanceof had some performance implications.

nicktindall: Also is it possible to deprecate the 2-param `setDefaultValue` seeing as `defaultValue` is ignored everywhere?

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -1622,6 +1699,22 @@ protected void setValue(Object o, ValueIn read, boolean overwrite) {
...
+                if (coll == null) {
+                    coll = collectionSupplier.get();
+                    unsafePutObject(o, offset, coll);
+                }
+                coll.clear();
+                return;
+            }
+
+            unsafePutObject(o, offset, defaultValue);
+        }
```
nicktindall: This seems pretty nasty, especially given collections are usually mutable, how does it compare to existing behaviour?

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -1796,6 +1910,22 @@ protected void setValue(Object o, ValueIn read, boolean overwrite) {
...
+                Map<?, ?> map = unsafeGetObject(o, offset);
+                if (map == null) {
+                    map = collectionSupplier.get();
+                    unsafePutObject(o, offset, map);
+                }
+                map.clear();
+                return;
+            }
+
+            unsafePutObject(o, offset, defaultValue);
```
nicktindall: Again this seems fairly dangerous. Interested to know how it compares to existing behaviour

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -1113,15 +1153,27 @@ protected void setValue(Object o, @NotNull ValueIn read, boolean overwrite) thro
...
-            bytes.writeUtf8(String.valueOf(field.get(o)));
+        protected void setDefaultValue(Object ignored, Object o) {
+            if (defaultValue != null && defaultValue instanceof Resettable && !(defaultValue instanceof DynamicEnum)) {
+                Object existingValue = unsafeGetObject(o, offset);
+
+                if (existingValue == defaultValue)
+                    return;
+
+                if (existingValue != null && existingValue.getClass() == defaultValue.getClass()) {
+                    ((Marshallable) existingValue).reset();
```
nicktindall: Could this end in an infinite loop with a self-referential DTO or e.g. a doubly-linked structure?

Not sure how you'd go about avoiding that in a garbage free way, but perhaps something to consider?

nicktindall (approved):
Looks mostly OK with some comments

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
JerryShea: Yes - the less instanceof the better. There can be some nasty and surprising performance implications

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
alamar: Yes, that's what the default implementation of `copy()` does, which is called in this method by default.

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
alamar: See above

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
alamar: Yes, it could; however, due to the safety of JVM, we'll just get a neat `StackOverflowError` out of it.

I don't think we could easily guard against this scenario. DTOs are supposed to be trees and not graphs, anyway, since it is assumed they are serialized eventually.

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
alamar: I've removed `setDefaultValue` for collections/maps since they were not doing anything that `copy()` won't do anyway.

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
alamar: `defaultValue` is used by default implementation of this method. I will keep it referenced.

PR #756, opened 2023-10-30
Remove Event.eventId().

yevgenp:


src/main/java/net/openhft/chronicle/wire/AbstractEventCfg.java
```diff
@@ -21,23 +21,11 @@
 import org.jetbrains.annotations.NotNull;
 
 public class AbstractEventCfg<E extends AbstractEventCfg<E>> extends AbstractMarshallableCfg implements Event<E> {
-    private String eventId = "";
+    private final String eventId = "";
```
maxim-ponomarev: Can we remove this field?

yevgenp: done

maxim-ponomarev (approved):

PR #769, opened 2023-11-08
Make reflection based proxies disable be default until removal in x.27

peter-lawrey:


src/test/java/net/openhft/chronicle/wire/UpdateInterceptorReturnTypeTest.java
```diff
@@ -53,15 +54,11 @@ static Wire createWire() {
     @Before
     public void setUp() {
         System.setProperty(DISABLE_WRITER_PROXY_CODEGEN, String.valueOf(disableProxyCodegen));
+        System.setProperty(DISABLE_PROXY_REFLECTION, String.valueOf(!disableProxyCodegen));
         if (disableProxyCodegen)
             expectException("Falling back to proxy method writer");
     }
 
-    @After
```
JerryShea: why remove?

JerryShea: ah I see

src/main/java/net/openhft/chronicle/wire/VanillaMethodReaderBuilder.java
```diff
@@ -138,10 +139,12 @@ private MethodReader createGeneratedInstance(Object... impls) {
                 }
             }
         } catch (Throwable e) {
+            if (Jvm.getBoolean(DISABLE_PROXY_REFLECTION, true))
+                throw Jvm.rethrow(e);
             classCache.put(fullClassName, COMPILE_FAILED);
             Jvm.warn().on(getClass(), "Failed to compile generated method reader - " +
```
JerryShea, line 222: this code is not exercised any more. We should parameterise at least one of the tests so that they exercise the proxy codegen still (in the same way as we do for method writers i.e. we have parameterised `UpdateInterceptorReturnTypeTest`)

JerryShea (approved):
Looks great but please consider adding a test as per the comment

peter-lawrey:
This appears to be out of date now with the x.26 roll

PR #781, opened 2023-12-05
Allow suppression of CNFE for interface fields

chroniclekevinpowe:
When class.not.found.for.missing.class.alias is set to false, no CNFE should be thrown for an interface field set to a missing class name. This allows the user to optionally suppress CNFE behaviour.


peter-lawrey (approved):
Can you change the name of the method to reflect the expected behaviour

Issue #738, opened 2023-10-16
JsonWire doesn't support typeLiterals


maxim-ponomarev:
`!type` declaration is not supported in JsonWire see pull request https://github.com/OpenHFT/Chronicle-Wire/pull/795

RobAustin:
should its show this "{\"implClass\":!type net.openhft.chronicle.wire.JSONWireTest}",

yevgenp:
> should its show this "{"implClass":!type net.openhft.chronicle.wire.JSONWireTest}",

That's not a valid JSON. Though my PR is still WiP. I can't get an expected output from @peter-lawrey  for all the tests above.

RobAustin:
I spoke with maxim today and ( I think ) he said this would be the best format for the UI:
````
{
  "@net.openhft.chronicle.wire.JSONWireTest$DtoWithClassReference": {
    "implClass": {
      "@type": "net.openhft.chronicle.wire.JSONWireTest"
    },
    "bool": "false"
  }
}
````

RobAustin:
@maxim-ponomarev told me that it was blocking him, so Ive put together a PR for review, https://github.com/OpenHFT/Chronicle-Wire/pull/795

RobAustin:
@maxim-ponomarev and @yevgenp  please just go ahead and alter the PR if you feel there should be changes to it.

RobAustin:
I see there are 5 test failures, I'll have to look at that later.

<img width="950" alt="image" src="https://github.com/OpenHFT/Chronicle-Wire/assets/6506056/96af4819-e39b-4f6e-b3a7-bf4f81f0674c">

RobAustin:
@Yevgen - please re-review

hft-team-city:
Released in [Chronicle-Wire-2.25ea5](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.25ea5), [BOM-2.25ea23](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.25ea23)

PR #774, opened 2023-11-17
Add class reference support for JSONWire.

yevgenp:
Fixes #738

PR #795, opened 2023-12-21
#738 added support for typeLiterals in json wire

RobAustin:


src/test/java/net/openhft/chronicle/wire/JSONWireTest.java
```diff
@@ -519,4 +520,30 @@ private static void readList(SimpleTwoLists record, List<String> data, ValueIn r
             }
         }
     }
-}
+
+    @Test
+    public void typeLiteral1() {
```
yevgenp, line 525: This test looks like subcase of the below one. Do we really need both?

src/main/java/net/openhft/chronicle/wire/JSONWire.java
```diff
@@ -725,6 +739,67 @@ public JSONWire rawText(CharSequence value) {
...
+        @NotNull
+        private Type consumeTypeLiteral() {
+            long start = bytes.readPosition();
+            consumePadding();
+            StringBuilder sb = Wires.acquireStringBuilderScoped().get();
+
+            int code = readCode();
+            if (code != '{') {
+                bytes.readPosition(start);
+                return UnresolvedType.of("DoesntExist");
```
yevgenp: "DoesntExist" looks like a hack to only satisfy WireTests#testLenientTypeLiteral.
We can't do that.

yevgenp (commented):
Unfortunately, this changes are not enough. There are few existing tests that use invalid JSON as expected outcome: AbstractUntypedFieldTest#typedFieldsShouldBeNonNullJsonWire, 
JSONEmptySequencesTest#emptySequence,
WireToOutputStreamTest#testNoSocket

Please see my PR https://github.com/OpenHFT/Chronicle-Wire/pull/774
I'm also concerned about backward compatibility.

maxim-ponomarev (approved):

yevgenp (approved):

PR #800, opened 2024-01-02
Add leading zeros support for SymbolsLongConverter.

### LINKED ISSUE #649: There's no way to encode a "zero" value in the SymbolsLongConverters

yevgenp:
Fixes #649

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/converter/SymbolsLongConverter.java
+++ b/src/main/java/net/openhft/chronicle/wire/converter/SymbolsLongConverter.java
@@ -62,7 +62,7 @@ public class SymbolsLongConverter implements LongConverter {
         for (int i = 0; i < decode.length; i++)
             encode[decode[i]] = (short) i;
 
-        maxParseLength = LongConverter.maxParseLength(length);
+        maxParseLength = LongConverter.maxParseLength(length) - 1;
     }
 
     @Override
@@ -82,6 +82,7 @@ public class SymbolsLongConverter implements LongConverter {
         lengthCheck(text);
 
         long v = 0;
+        v += text.length();
         for (int i = 0; i < text.length(); i++) {
             final char ch = text.charAt(i);
 
@@ -118,6 +119,10 @@ public class SymbolsLongConverter implements LongConverter {
             text.append(decode[v]);
         }
 
+        if (text.length() > 1) {
+            text.deleteCharAt(text.length() - 1);//remove optional (backward compatible) length
+        }
+
         StringUtils.reverse(text, start); // Reverse the result since it's constructed backward.
 
         if (text.length() > start + maxParseLength()) {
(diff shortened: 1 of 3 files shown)


yevgenp:
It's a possible implementation option which comes with its' own drawbacks (maxLength -1, long overflow).

nicktindall:
Very clever. I like this approach better than https://github.com/OpenHFT/Chronicle-Wire/pull/822 because the latter still strips leading zeroes. If everyone's happy with the reduced max length, and we're certain that it's backwardly compatible then it's approved by me.

nicktindall (approved):

nicktindall:
Be good to get @peter-k-lawrey's thoughts on this one

JerryShea:
> Be good to get @peter-k-lawrey's thoughts on this one

What if he disagrees with @peter-lawrey ?

nicktindall:
> > Be good to get @peter-k-lawrey's thoughts on this one
> 
> What if he disagrees with @peter-lawrey ?

Is there a hierarchy of Peters? surely they don't always agree

JerryShea:
> > > Be good to get @peter-k-lawrey's thoughts on this one
> > 
> > 
> > What if he disagrees with @peter-lawrey ?
> 
> Is there a hierarchy of Peters? surely they don't always agree

Ask @peter-lawrey-admin - he surely is the boss

src/main/java/net/openhft/chronicle/wire/converter/SymbolsLongConverter.java
```diff
@@ -149,6 +154,9 @@ public void append(Bytes<?> text, long value) {
             text.append(decode[v]);
         }
 
+        if (!text.isEmpty())
+            text.unwrite(text.writePosition() - 1, 1);
+
```
nicktindall, line 159: Actually this isn't backward or forward compatible. I really like it though, perhaps we make it a separate implementation and deprecate this one....

e.g. when encoded by the new impl and read by the old impl you get
`5HELLO`

and when encoded by the old impl and read by the new impl you get:
`ELLO`

nicktindall (changes requested):
We should make this a separate implementation and deprecate the existing implementation, or even keep them as alternatives. It's not forward or backward coimpatible which is an issue if you have persisted data, or if you upgrade services that communicate independently.

yevgenp:
Already implemented in #822

PR #805, opened 2024-01-04
Deprecate StringUtils.setCount as not compatible with Java 21.

yevgenp:


peter-lawrey (approved):
if we can call setLength, we should always do so.

PR #811, opened 2024-01-17
Fix CSVWireTest. Deprecate unused methods.

### LINKED ISSUE #804: Fix CSVWireTest

yevgenp:
Closes #804

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireType.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireType.java
@@ -36,7 +36,6 @@ import java.io.IOException;
 import java.io.Serializable;
 import java.lang.reflect.Constructor;
 import java.net.URL;
-import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Spliterator;
 import java.util.Spliterators;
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/wire/WireType.java
```diff
@@ -600,47 +599,6 @@ public boolean tryAdvance(@NotNull Consumer<? super T> action) {
                 }, false);
     }
 
-    @NotNull
-    public <T> Map<String, T> fromFileAsMap(String filename, @NotNull Class<T> tClass) throws IOException, InvalidMarshallableException {
```
JerryShea, line 604: should deprecate before removing

src/test/java/net/openhft/chronicle/wire/CSVWireTest.java
```diff
@@ -19,15 +19,10 @@
...
 import org.junit.Assert;
-import org.junit.Ignore;
 import org.junit.Test;
 
-import java.io.IOException;
-import java.util.Map;
-
 import static org.junit.Assert.*;
 
-@Ignore("TODO FIX")
```
JerryShea, line 30: nice

src/test/java/net/openhft/chronicle/wire/CSVWireTest.java
```diff
@@ -62,33 +59,6 @@ public void tstFrom2() {
         doTestWire(wire);
     }
 
-    @Test
-    public void tstFrom3() throws IOException {
```
JerryShea, line 66: why remove?

src/main/java/net/openhft/chronicle/wire/WireType.java
yevgenp: This method is not used anywhere (across all Chronicle repos) but in the following removed test. 
IMO it does not make sense to spend more time to fix it (I've tried) as it's a dead code.

JerryShea: OK. @peter-lawrey 's decision.... I am OK to remove although if we want to preserve API compatibility we could deprecate but replace the contents with `throw new UnsupportedOperationException()`

peter-lawrey:
Unless we know a feature is broken, we should delay removal for 2 versions. i.e. removal in x.27.

yevgenp:
> Unless we know a feature is broken, we should delay removal for 2 versions. i.e. removal in x.27.

Like I've mentioned above that method is not used anywhere (across all Chronicle repos) but in the following removed test.
It'd require quite an effort to make it work. Which makes no sense IMO.
Do you suggest to deprecate it?

src/main/java/net/openhft/chronicle/wire/WireType.java
peter-lawrey: I wouldn't throw UOE; it's better to fail at compile time.

JerryShea: so @peter-lawrey you would rather remove the methods?

peter-lawrey: All to be deprecated for removal in x.27

yevgenp:
Removed broken method and it's test. Deprecated two others.

peter-lawrey (approved):

Issue #733, opened 2023-10-09
Support marshalling of Java 14 records


alamar:
Right now, attempting do to so results in

> UnsupportedOperationException: can't get field offset on a record class: private final xxx..dto.Pathway xxx.dto.ClientIdentifierDto.pathway

StevenCurran:
+1 this would be very useful!

peter-lawrey:
In the short term we should produce a meaningful error message until this is properly supported.
- we should explore how this can be supported.
- whether object reuse is possible or desirable to reduce object allocations.

PR #814, opened 2024-01-18
Throw UnsupportedOperationException on record class.

yevgenp:
Temporary workaround for #733

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
@@ -47,12 +47,22 @@ import static net.openhft.chronicle.core.UnsafeMemory.*;
 public class WireMarshaller<T> {
     private static final Class[] UNEXPECTED_FIELDS_PARAMETER_TYPES = {Object.class, ValueIn.class};
     private static final FieldAccess[] NO_FIELDS = {};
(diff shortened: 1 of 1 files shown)


src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -817,6 +817,8 @@ public static Object create(@NotNull Field field, @Nullable Object defaultObject
                     case "net.openhft.chronicle.bytes.Bytes":
                         return new BytesFieldAccess(field);
                     default:
+                        if (Jvm.isJava14Plus() && (boolean) Jvm.getMethod(Class.class, "isRecord").invoke(type))
```
peter-lawrey: Can this be cached
static {
Method isRecord = null;
try {
   if (Jvm.isJava14Plus())
       isRecord = Jvm.getMethod(Class.class, "isRecord");
} catch (Exception expected) {
   isRecord = null; 
}
IS_RECORD = isRecord;
}
// later
if (IS_RECORD != null && IS_RECORD.invoke(type))

yevgenp: Done

peter-lawrey (approved):

PR #820, opened 2024-01-30
Roll to new parent pom

tgd:
Roll to new parent pom to avoid breakages from NotNull plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
     <parent>
         <groupId>net.openhft</groupId>
         <artifactId>java-parent-pom</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -22,7 +22,7 @@
     <parent>
         <groupId>net.openhft</groupId>
         <artifactId>java-parent-pom</artifactId>
-        <version>1.25.0</version>
+        <version>1.25.1</version>
```
yevgenp, line 25: There is 1.25.**2** already

yevgenp (approved):

PR #822, opened 2024-01-31
Added a @ShortText specifically designed to encode short strings.

peter-lawrey:
`@Base85` is designed for encoding numbers as short strings.
This is designed to encode short strings including leading zeros as long.

Closes https://github.com/OpenHFT/Chronicle-Wire/issues/649

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/ShortTextLongConverter.java
+++ b/src/main/java/net/openhft/chronicle/wire/ShortTextLongConverter.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2016-2020 chronicle.software
+ *
+ *       https://chronicle.software
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.openhft.chronicle.wire;
+
+/**
+ * Implements a Base 85 encoding scheme specifically designed for converting short text strings into long values.
+ * <p>
+ * The ShortTextLongConverter is optimized for encoding text strings of up to 10 characters. It efficiently handles
+ * leading spaces by truncating them, similar to how {@link Base85LongConverter} truncates leading zeros. This feature
+ * is particularly useful in scenarios where text padding can be ignored.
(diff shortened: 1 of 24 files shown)


src/main/java/net/openhft/chronicle/wire/ShortTextLongConverter.java
```diff
@@ -0,0 +1,90 @@
...
+     * Specifies the custom character set used for the Base85 encoding in this implementation.
+     * The character set is thoughtfully chosen to include a diverse range of symbols, letters,
+     * and digits, facilitating a robust and flexible encoding process.
+     */
+    private static final String CHARS = " " +
+            "123456789" +
+            ":;<=>?@" +
+            "ABCDEFGHIJKLMNOPQRSTUVWXYZ_" +
+            "abcdefghijklmnopqrstuvwxyz" +
+            "\"#$%&'()*+,-./0";
```
yevgenp, line 61: I'd move 0 to the very start of the constant. It took me some time to find it :)

src/test/java/net/openhft/chronicle/wire/utils/dto/AbstractEvent.java
```diff
@@ -5,13 +5,13 @@
 package net.openhft.chronicle.wire.utils.dto;
 
 import net.openhft.chronicle.wire.*;
-import net.openhft.chronicle.wire.converter.Base85;
 import net.openhft.chronicle.wire.converter.NanoTime;
+import net.openhft.chronicle.wire.converter.ShortText;
 
 public class AbstractEvent<E extends AbstractEvent<E>> extends SelfDescribingMarshallable {
-    @Base85
+    @ShortText
```
yevgenp, line 12: Does it make sense to replace `@Base85` with `@ShortText` on any fields except `String` ?
We're not deprecating `@Base85`, right?

yevgenp (approved):

src/main/java/net/openhft/chronicle/wire/ShortTextLongConverter.java
peter-lawrey: The first character is the one which is ignored if it is leading. 0 specifically can't be the first or it will be truncated

src/test/java/net/openhft/chronicle/wire/utils/dto/AbstractEvent.java
peter-lawrey: We aren't deprecating Base85 yet, however we might in the future.

src/main/java/net/openhft/chronicle/wire/ShortTextLongConverter.java
yevgenp: So ShortTextLongConverter ignores leading space, correct?

nicktindall:
Should update demos to use this now https://github.com/search?q=%40Base85+repo%3AOpenHFT%2FChronicle-Queue-Demo&type=code

PR #823, opened 2024-01-31
Support generateTuple being set for an individual WireIn, rather than global all or nothing.

peter-lawrey:
Tuples support reading data for a type you don't have locally and then passing it on by writing it.


yevgenp (approved):

RobAustin (approved):

maxim-ponomarev (changes requested):
WireTypeConverter still doesn't handle unknown classes: see `WireTypeConverterWithTuplesTest` in [638a7e1](https://github.com/OpenHFT/Chronicle-Wire/pull/823/commits/638a7e19d7e5de299dc4adb7ca180afe4346c0ff)

maxim-ponomarev:
Added test (see above)

maxim-ponomarev:
Let me know, if I'm misusing Tuples and they are not supposed to retain the type information.

peter-lawrey:
Replaced by a newer PR

PR #831, opened 2024-02-09
If compiled with Java 11, still work on Java 8.

peter-lawrey:


tgd (approved):

JerryShea:
Change looks good but it may be worth applying `maven.compiler.release` across the board to all projects - see https://github.com/OpenHFT/Chronicle-Network/commit/b0276618b510e83886766cb8c896a0870c0fa024

PR #836, opened 2024-02-22
Add subSequence parse to LongConverter - closes #835

### LINKED ISSUE #835: Add sub-sequence parse() to LongConverter
> parseSubSequence(CharSequence sequence, int from, int to)

To avoid creating extra garbage when working with partially parsed input.

alamar:


src/test/java/net/openhft/chronicle/wire/Base85LongConverterTest.java
```diff
@@ -61,6 +61,25 @@ public void parse() {
         }
     }
 
+    @Test
```
tgd, line 64: Please could you enhance the names of the test methods from `parseX` - I am not sure what some of them are testing

src/main/java/net/openhft/chronicle/wire/LongConverter.java
```diff
@@ -141,6 +155,19 @@ default void lengthCheck(CharSequence text) {
...
     }
 
+    /**
+     * Checks that the length of the provided text does not exceed the allowable maximum.
+     *
+     * @param beginIndex the beginning index, inclusive.
+     * @param endIndex the ending index, exclusive.
+     * @throws IllegalArgumentException if the text length exceeds the maximum allowable length.
+     */
+    default void lengthCheck(int beginIndex, int endIndex) {
```
tgd: Nice! Please add some explicit test coverage for the bounds checking if we don't have it already

src/main/java/net/openhft/chronicle/wire/IdentifierLongConverter.java
```diff
@@ -54,11 +54,26 @@ protected IdentifierLongConverter() {
...
     public long parse(CharSequence text) {
-        return text.length() <= 10
-                ? SMALL_POSITIVE.parse(text)
-                : NanoTimestampLongConverter.INSTANCE.parse(text);
+        return parse(text, 0, text.length());
     }
 
+    /**
+     * Parses a part of the provided {@link CharSequence} into a long identifier.
+     * The parsing behavior changes depending on the length of the CharSequence.
```
tgd: > The parsing behavior changes depending on the length of the CharSequence.

I think this javadoc needs to be expanded - how does the behaviour change?

tgd: Also can add a note that it throws IllegalArgumentException for invalid indices

tgd (changes requested):
Nice one Ilya - I have requested a few small changes. Also please get one person from APAC to review this prior to merge.

Please can you add an appropriate entry in the Wire README giving an idea of when this would be used, why and how to use it.

Thanks

tgd:
Also Ilya - heads up the PR is targeting `ea` not `develop`

JerryShea (approved):
No further comments!

src/test/java/net/openhft/chronicle/wire/Base32LongConverterTest.java
```diff
@@ -33,6 +33,25 @@ public void parse() {
...
     }
 
+    @Test
+    public void parse2() {
+        LongConverter c = Base32LongConverter.INSTANCE;
+        // System.out.println(c.asString(-1L));
+        String s = ",O,A,L,ZZ,QQ,ABCDEGHIJKLM,5OPQRSTVWXYZ,JZZZZZZZZZZZ,";
+        int oldPos = 0;
+        int newPos;
+        int comparsions = 9;
```
yevgenp: typo?

src/test/java/net/openhft/chronicle/wire/Base32LongConverterTest.java
```diff
@@ -33,6 +33,25 @@ public void parse() {
...
+        while ((newPos = s.indexOf(',', oldPos)) >= 0) {
+            long v = c.parse(s, oldPos, newPos);
+            StringBuilder sb = new StringBuilder();
+            c.append(sb, v);
+            assertEquals(s.substring(oldPos, newPos), sb.toString());
+            oldPos = newPos + 1;
+            comparsions--;
+        }
+        assertEquals(0, comparsions);
```
yevgenp: Not sure if the test has to be so complicated.

src/main/java/net/openhft/chronicle/wire/LongConverter.java
```diff
@@ -67,6 +67,20 @@ static int maxParseLength(int based) {
...
+     * {@code long} primitive.
+     *
+     * @param text character sequence containing the string representation of the value.
+     * @param beginIndex the beginning index, inclusive.
+     * @param endIndex the ending index, exclusive.
+     *
+     * @return the parsed {@code text} as an {@code long} primitive.
+     */
+    default long parse(CharSequence text, int beginIndex, int endIndex) {
+        return parse(text.toString().substring(beginIndex, endIndex));
```
yevgenp, line 86: It seems that the original idea was to avoid extra object allocation (GC) but `text.toString()` is doing the opposite.

alamar: That's true, but this is just a default implementation, which is supposed to be overriden and then provide more efficient approach. Should probably be noted in javadoc.

src/test/java/net/openhft/chronicle/wire/Base32LongConverterTest.java
alamar: Why? It would fail if the value does not match the outcome

src/test/java/net/openhft/chronicle/wire/Base32LongConverterTest.java
alamar: I will try to extract this logic but overall it makes sense in my opinion

src/main/java/net/openhft/chronicle/wire/LongConverter.java
alamar: I've added a javadoc about it

alamar (commented):
please re-review @yevgenp @tgd

src/test/java/net/openhft/chronicle/wire/Base32LongConverterTest.java
yevgenp: compar**IS**ons

yevgenp (approved):

tgd (approved):
Approved - thanks Ilya

src/test/java/net/openhft/chronicle/wire/Base32LongConverterTest.java
alamar: This is now also fixed

alamar:
#835

Issue #841, opened 2024-02-28
Flatten the call stack for Method Readers


peter-lawrey:

peter-lawrey:
Fixed in PR #842

hft-team-city:
Released in [Chronicle-Wire-2.25ea10](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.25ea10), [BOM-2.25ea30](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.25ea30)

PR #842, opened 2024-02-28
Flatten the call stack for Method Readers #841

peter-lawrey:


tgd (approved):

tgd:
Hi Peter - can you share some background on this one?

PR #847, opened 2024-02-29
Support copying between yaml and json

peter-lawrey:


src/main/java/net/openhft/chronicle/wire/JSONWire.java
```diff
@@ -619,12 +621,21 @@ public JSONValueOut typePrefix(@NotNull CharSequence typeName) {
...
                 bytes.append("\"@");
-                bytes.append(typeName);
+                bytes.append(applyAsAlias(classLookup, typeName));
                 bytes.append("\":");
             }
             return this;
         }
 
+        private CharSequence applyAsAlias(ClassLookup classLookup, CharSequence typeName) {
+            // TODO use classLookup.applyAsAlias(typeName);
```
tgd, line 631: Hi @peter-lawrey - there is still a TODO here

maxim-ponomarev (approved):

src/main/java/net/openhft/chronicle/wire/JSONWire.java
peter-lawrey: Indeed, once the changes in Core have propagated this method can be used instead.

Issue #851, opened 2024-03-14
Improve efficiency of writing MessageHistory as bytes


peter-lawrey:
Currently, VanillaMessageHistory uses marshalling code via Bytes; however, it's faster to use raw memory access when direct memory is available.

JerryShea:
@peter-lawrey I believe this can be closed?

PR #856, opened 2024-03-15
Make history.as.bytes true by default to improve efficiency of writing MessageHistory as bytes #851

peter-lawrey:
Writing the `MessageHistory` as a `SelfDescribingMarshallable` showed up as significant contributor in a recent benchmark.
This makes writing the direct bytes format, the default which can be turned of using the `history.self.describing` property
Regardless of this setting, either format is read as it was written.
The estimated impact for even a short history is typically around 0.5 us, and 1.1 us on the 99.9%ile.


tgd (approved):

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
```diff
@@ -439,6 +444,14 @@ private void copyOne(@NotNull WireOut wire, boolean first) throws InvalidMarshal
         }
     }
 
+    private void copyHistoryMessage(@NotNull WireOut wire) {
+        VanillaMessageHistory vmh = new VanillaMessageHistory();
```
tgd: Does this need to be pooled or pulled into a field reference and recycled?

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
```diff
@@ -810,6 +831,11 @@ private StringBuilder readSpecialField(int peekCode, CharSequence keyName, int k
             case FIELD_NUMBER:
                 bytes.uncheckedReadSkipOne();
                 long fieldId = bytes.readStopBit();
+                if (fieldId == MethodReader.MESSAGE_HISTORY_METHOD_ID) {
+                    sb.setLength(0);
+                    sb.append(MethodReader.HISTORY);
```
JerryShea, line 1164: why is this required? generated method readers handle MESSAGE_HISTORY_METHOD_ID already

src/test/java/net/openhft/chronicle/wire/MessageHistoryTest.java
```diff
@@ -77,8 +77,8 @@ public void checkSerialiseBytes() {
         history2.readMarshallable(bw);
         assertEquals("VanillaMessageHistory { " +
                 "sources: [1=0xff,2=0xfff], " +
-                "timings: [1000000000000000000,1000000000000010000,120962203520100], " +
-                "addSourceDetails=false }", history2.toString());
+                "timings: [1000000000000000000,1000000000000010000,120962203520100,120962203520100], " +
```
JerryShea, line 104: why different?

src/test/java/net/openhft/chronicle/wire/channel/echo/RecordHistoryEchoHandlerTest.java
```diff
@@ -41,6 +41,9 @@ public class RecordHistoryEchoHandlerTest extends WireTestCommon {
     private static void doTest(ChronicleContext context, ChannelHandler handler) {
         ChronicleChannel channel = context.newChannelSupplier(handler).connectionTimeoutSecs(1).get();
         assertTrue(channel.recordHistory());
+        VanillaMessageHistory history = (VanillaMessageHistory) MessageHistory.get();
+        history.reset(1, 128);
+//        history.historyMethodId(false);
```
JerryShea: pls remove commented-out code

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
```diff
@@ -460,7 +473,13 @@ public void readWithLength(@NotNull WireOut wire, int len) throws InvalidMarshal
         try {
             bytes.readLimit(newLimit);
             @NotNull final ValueOut wireValueOut = wire.getValueOut();
-            switch (getBracketTypeNext()) {
+            BracketType bracketType = getBracketTypeNext();
+            switch (bracketType) {
+                case UNKNOWN:
+                    bytes.uncheckedReadSkipOne();
+                    copyHistoryMessage(wire);
+                    return;
```
nicktindall, line 655: Is it right to use `UNKNOWN` to convey `HISTORY_MESSAGE` here? seems a bit obfuscated to me.

src/main/java/net/openhft/chronicle/wire/channel/ChronicleChannel.java
```diff
@@ -88,7 +88,8 @@ default <T> T readOne(StringBuilder eventType, Class<T> expectedType) throws Clo
             try (DocumentContext dc = readingDocument()) {
                 if (dc.isPresent()) {
                     ValueIn in = dc.wire().read(eventType);
-                    if (StringUtils.isEqual(eventType, MethodReader.HISTORY)) {
+                    if (StringUtils.isEqual(eventType, MethodReader.HISTORY) ||
+                            StringUtils.isEqual(eventType, "" + MethodReader.MESSAGE_HISTORY_METHOD_ID)) {
```
nicktindall, line 92: I assume this gets optimised into a constant by the compiler, if not perhaps we should make it a constant?

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
JerryShea: Agree

tgd (changes requested):
Hi Peter - flipping the status on this to request changes as there are a few comments

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
peter-lawrey: The problem is that this code is widely used and the HISTORY_MESSAGE only applies here. The alternative is to create a code just for this purpose, or add an option that is ignored or produces an error in the cases where it doesn't apply.

src/main/java/net/openhft/chronicle/wire/channel/ChronicleChannel.java
peter-lawrey: It should as it's known at compile time.

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
peter-lawrey: I have added an enum as it improves clarity without as much overhead as I feared.

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
peter-lawrey: MethodReaders will recognise the field numbers however many tools don't use generated method readers. e.g. Chronicle Reader

src/test/java/net/openhft/chronicle/wire/MessageHistoryTest.java
peter-lawrey: Added addSourceDetails(true); to SetTimeMessageHistory for the benefit of other tests

src/test/java/net/openhft/chronicle/wire/MessageHistoryTest.java
```diff
@@ -77,8 +77,8 @@ public void checkSerialiseBytes() {
...
+                "addSourceDetails=true }", history2.toString());
```
peter-lawrey, line 105: @JerryShea printed here

src/test/java/net/openhft/chronicle/wire/channel/echo/RecordHistoryEchoHandlerTest.java
peter-lawrey: done

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
JerryShea: Unless you give it `-m` argument. But point taken

JerryShea: Could we add a comment explaining why?

src/main/java/net/openhft/chronicle/wire/BinaryWireCode.java
```diff
@@ -35,7 +35,7 @@ public enum BinaryWireCode {
     // sequence of length 0 - 255
 //        public static final int BYTES_LENGTH64 = 0x83;
 
-    public static final int BYTES_MARSHALLABLE = 0x86; // explicitly BytesMarshallable.
+    public static final int HISTORY_MESSAGE = 0x86; // was BYTES_MARSHALLABLE, but only used for this purpose
```
JerryShea: we should probably go back in time and mark `BYTES_MARSHALLABLE` as deprecated? Or are you confident enough that it is a internal or domestic API and won't be likely to be used outside of Chronicle products?

JerryShea (approved):

nicktindall (approved):

src/main/java/net/openhft/chronicle/wire/BinaryWireCode.java
peter-lawrey: Harmless to leave as Deprecated for removal.

tgd (approved):

RobAustin (approved):

PR #862, opened 2024-03-22
Error on deprecated

peter-lawrey:


pom.xml
```diff
@@ -450,6 +448,15 @@
     </scm>
 
     <profiles>
+        <profile>
+            <id>java11-release8</id>
```
tgd, line 452: Do we need this bit?

src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -77,8 +77,9 @@ public enum Wires {
     public static final int SPB_HEADER_SIZE = 4;
     public static final List<Function<Class, SerializationStrategy>> CLASS_STRATEGY_FUNCTIONS = new CopyOnWriteArrayList<>();
 
-    @Deprecated(/* for removal in x.26 */)
+    // @Deprecated(/* for removal in x.26 */)
```
tgd, line 81: If de-deprecating let's delete or bump deprecation to future version

src/main/java/net/openhft/chronicle/wire/channel/ChronicleChannelCfg.java
```diff
@@ -74,9 +74,9 @@ public Set<HostPortCfg> hostPorts() {
      * @param port     the port number
      * @return this configuration instance
      */
-    public ChronicleChannelCfg<C> addHostnamePort(String hostname, int port) {
+    public C addHostnamePort(String hostname, int port) {
```
tgd, line 77: I wonder if this signature change will break any callsites, if so we may need to put this signature through a deprecation cycle before changing it

tgd (changes requested):
Hi Peter looking good - I added a few comments highlighting potential changes

pom.xml
peter-lawrey: Yes. Without it we get a warning which prevents the build

src/main/java/net/openhft/chronicle/wire/Wires.java
peter-lawrey: Added a comment as to why this is commented out. Not de-deprecating it

src/main/java/net/openhft/chronicle/wire/channel/ChronicleChannelCfg.java
peter-lawrey: It's only possible to do that by adding a new method. This functionality is not widely used.

pom.xml
yevgenp: I think the best solution would be https://github.com/OpenHFT/OpenHFT/pull/222/files. 
Also it won't prob. work on Java21+

src/main/java/net/openhft/chronicle/wire/channel/ChronicleChannelCfg.java
yevgenp: It does not seem right. Why not simply return raw ChronicleChannelCfg? 
I don't see any parametrization used in the method.

src/main/java/net/openhft/chronicle/wire/channel/ChronicleContext.java
```diff
@@ -207,8 +207,7 @@ public ChronicleChannelSupplier newChannelSupplier(ChannelHandler handler) throw
 
         connectionSupplier
                 .protocol(url().getProtocol())
-                .hostname(hostname == null || hostname.isEmpty() ? "localhost" : hostname)
-                .port(port)
+                .addHostnamePort(hostname == null || hostname.isEmpty() ? "localhost" : hostname, port)
```
yevgenp, line 210: Why? That's not a common use for Builder pattern.

pom.xml
peter-lawrey: This works on Java 21 as I added -Xlint:deprecation,-options turns off the warning that Java 8 might not be supported in the future.

src/main/java/net/openhft/chronicle/wire/channel/ChronicleChannelCfg.java
peter-lawrey: It's needed so that subclasses have the right co-variant return type.

src/main/java/net/openhft/chronicle/wire/channel/ChronicleContext.java
peter-lawrey: It's to add a pair of hostnames/ports, which makes more sense when there are multiple hostnames/ports to be added.

yevgenp (approved):

tgd (approved):

PR #868, opened 2024-04-03
X.26 lint all

peter-lawrey:


src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
```diff
@@ -404,13 +404,14 @@ public String toString() {
...
+    @SuppressWarnings("unchecked")
     @Override
-    public @NotNull VanillaMessageHistory deepCopy() throws InvalidMarshallableException {
+    public @NotNull <T> T deepCopy() throws InvalidMarshallableException {
         @NotNull VanillaMessageHistory copy = super.deepCopy();
         // remove the extra timing
         copy.timingsArray[this.timings] = 0;
         copy.timings = this.timings;
-        return copy;
+        return (T) copy;
```
yevgenp: I don't think this change is valid or makes sense: you're parameterizing method return value but in fact always return `VanillaMessageHistory`. It's confusing.

yevgenp (approved):

src/main/java/net/openhft/chronicle/wire/VanillaMessageHistory.java
peter-lawrey: I agree, however in Java 21 it only compiles without warning with this combination.

peter-lawrey: deepCopy() in net.openhft.chronicle.wire.VanillaMessageHistory implements <T>deepCopy() in net.openhft.chronicle.wire.Marshallable
  return type requires unchecked conversion from @org.jetbrains.annotations.NotNull net.openhft.chronicle.wire.VanillaMessageHistory to T

PR #874, opened 2024-04-16
Added test demonstrating sb not clearing after TextWire#readLength

chroniclekevinpowe:
It appears that TextWire#readLength (called as part of converting a ValueIn to text) has a side effect of leaving the string buffer set to the last read value, and that value is then prepended to the converted text.

### DIFF
--- a/src/test/java/net/openhft/chronicle/wire/marshallable/MarshallingJSONStringTest.java
+++ b/src/test/java/net/openhft/chronicle/wire/marshallable/MarshallingJSONStringTest.java
@@ -1,0 +1,43 @@
+package net.openhft.chronicle.wire.marshallable;
+
+import net.openhft.chronicle.core.io.IORuntimeException;
+import net.openhft.chronicle.core.io.InvalidMarshallableException;
+import net.openhft.chronicle.wire.Marshallable;
+import net.openhft.chronicle.wire.WireIn;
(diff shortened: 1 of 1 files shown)


src/test/java/net/openhft/chronicle/wire/marshallable/MarshallingJSONStringTest.java
```diff
@@ -0,0 +1,43 @@
...
+                "  config: {\n" +
+                "    \"username\": \"sampleApp\",\n" +
+                "    \"password\": \"samplePassword\",\n" +
+                "    \"publishPort\": 4021,\n" +
+                "    \"subscribePort\": 4024,\n" +
+                "  }\n" +
+                "}";
+
+        MarshallingJSONStringTest read = Marshallable.fromString(configJson);
+        assertFalse(read.configAsJSON.startsWith("4024"));
```
nicktindall, line 40: Should we do

`assertEquals("expected", read.configAsJSON)` instead, just so we get a better failure message?

nicktindall: Also to catch any other garbling

chroniclekevinpowe: Good point - done!

PR #875, opened 2024-04-16
Added exploratory fix for MarshallingJSONStringTest

chroniclekevinpowe:
StringBuffer length is now reset at end of readLength, clearing StringBuffer.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/TextWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/TextWire.java
@@ -1670,6 +1670,8 @@ public class TextWire extends YamlWireOut<TextWire> {
                 // Reset the consumption flag and reading position
                 this.consumeAny = false;
                 bytes.readPosition(start);
+                // @TODO - use ScopedResource<StringBuilder> for consistency throughout YamlWireOut - https://github.com/OpenHFT/Chronicle-Wire/issues/879
+                sb.setLength(0);
             }
         }
 
--- a/src/test/java/net/openhft/chronicle/wire/marshallable/MarshallingJSONStringTest.java
+++ b/src/test/java/net/openhft/chronicle/wire/marshallable/MarshallingJSONStringTest.java
@@ -1,0 +1,43 @@
+package net.openhft.chronicle.wire.marshallable;
+
+import net.openhft.chronicle.core.io.IORuntimeException;
+import net.openhft.chronicle.core.io.InvalidMarshallableException;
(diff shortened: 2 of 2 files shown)


src/main/java/net/openhft/chronicle/wire/TextWire.java
```diff
@@ -1670,6 +1670,7 @@ protected long readLengthMarshallable() {
                 // Reset the consumption flag and reading position
                 this.consumeAny = false;
                 bytes.readPosition(start);
+                sb.setLength(0);
```
nicktindall, line 1674: It looks like the pattern is usually to call `acquireStringBuilder()` which will set the length to zero before returning it to you. Perhaps the problem is someone using `sb` directly rather than attaining it via `acquireStringBuilder()`

A better/safer/more idiot-proof pattern would be to use `ScopedThreadLocal<StringBuilder>` but I don't fancy doing that refactor.

chroniclekevinpowe: In this case, acquireStringBuilder() is called as the first line of TextWire#text - the readLength method then uses the underlying StringBuilder as part of consuming content to determine length.

So if I'm understanding correctly, that pattern has been observed, but too early to avoid the impact from readLength.

nicktindall: This is actually a prime example of where `ScopedThreadLocal<StringBuilder>` would eliminate the possibility of/alert you to the presence of a bug. The nested context is garbling a resource that the outer context thinks it has "acquired".

I kinda think this change will eliminate a single instance of a bug that's probably repeated all through this class.

nicktindall: My vote would be to investigate the use of `ScopedThreadLocal<StringBuilder>` in place of the `sb` field.

chroniclekevinpowe: Can absolutely see where you're coming from, @nicktindall - appreciated!

JerryShea (approved):

src/main/java/net/openhft/chronicle/wire/TextWire.java
JerryShea: Lets merge this fix but please add a TODO: on that line for ScopedThreadLocal so we can remove that line

chroniclekevinpowe: Done!

PR #877, opened 2024-04-18
Explore safey of YamlWireOut#sb re-use

nicktindall:
I ignored the warning that indicates potentially unsafe re-use but there's a lot of them

They're not necessarily unsafe, it just indicates where sb is use in an outer scope then again in an inner scope. It could be that the outer scope doesn't care that the buffer gets trashed. Or it could be an issue like we've seen.

There are probably lots of "problematic" re-use patterns that look like this:

```java

void doSomething() {
    //...
   try (ScopedResource<StringBuilder> sbR = acquireStringBuilder()) {
      StringBuilder sb = sbR.get();
      readFieldName(sb);
      if (sb.contentEquals("something")) {
         doSomethingElse():
      }
   }
}

void doSomethingElse() {
   try (ScopedResource<StringBuilder> sbR = acquireStringBuilder()) {
      StringBuilder sb = sbR.get();
      //...
   }
}
```

Which is actually not problematic, it's a false-positive to say that the above represents unsafe re-use. The way to solve that is to tighten up the outer scope e.g.

```java
void doSomething() {
   //...
   if (fieldNameIs("something")) {
      doSomethingElse():
   }
}

void doSomethingElse() {
   try (ScopedResource<StringBuilder> sbR = acquireStringBuilder()) {
      StringBuilder sb = sbR.get();
      //...
   }
}

boolean fieldNameIs(String fieldName) {
   try (ScopedResource<StringBuilder> sbR = acquireStringBuilder()) {
      StringBuilder sb = sbR.get();
      readFieldName(sb);
      return sb.contentEquals(fieldName);
   }
}

```

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/YamlWireOut.java
+++ b/src/main/java/net/openhft/chronicle/wire/YamlWireOut.java
@@ -32,6 +32,9 @@ import net.openhft.chronicle.core.io.IORuntimeException;
 import net.openhft.chronicle.core.io.IOTools;
 import net.openhft.chronicle.core.io.InvalidMarshallableException;
 import net.openhft.chronicle.core.pool.ClassLookup;
+import net.openhft.chronicle.core.pool.StringBuilderPool;
+import net.openhft.chronicle.core.scoped.ScopedResource;
+import net.openhft.chronicle.core.scoped.ScopedResourcePool;
 import net.openhft.chronicle.core.values.*;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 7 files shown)


src/test/java/net/openhft/chronicle/wire/WireTestCommon.java
```diff
@@ -90,6 +90,12 @@ public void recordExceptions() {
         exceptions = Jvm.recordExceptions();
     }
 
+    @Before
+    public void ignorePoolCapacityExceededErrors() {
+        // Remove this to see where we have problematic re-use of YamlWire#acquireStringBuffer
+        ignoreException("Pool capacity exceeded, consider increasing maxInstances");
```
nicktindall, line 96: Get rid of this ignore to see where we might have an issue

src/main/java/net/openhft/chronicle/wire/YamlWireOut.java
```diff
@@ -85,7 +88,7 @@ public abstract class YamlWireOut<T extends YamlWireOut<T>> extends AbstractWire
     }
 
     protected final YamlValueOut valueOut = createValueOut();
-    protected final StringBuilder sb = new StringBuilder();
+    protected final ScopedResourcePool<StringBuilder> sb = StringBuilderPool.createThreadLocal(1);
```
nicktindall, line 91: Note that this probably doesn't need to be a thread-local given that YamlWireOut is single-threaded anyway. Making a single-threaded implementation of ScopedResourcePool is trivial.

Issue #886, opened 2024-04-26
JSONWire to parse doubles/ints surrounded by quotes


JerryShea:
if a DTO has a double field "price" then parsing in this JSON results in price's value == -0.0
```
"price": "1.234"
```
If you take away the quotes it is parsed correctly as 1.234 e.g
```
"price": 1.234
```

peter-lawrey:
Added double/float/long/int with single and double quote.

hft-team-city:
Released in [Chronicle-Wire-2.25ea14](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.25ea14), [BOM-2.25ea52](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.25ea52)

PR #887, opened 2024-04-26
JSONWire to parse doubles/ints surrounded by quotes #886

peter-lawrey:


JerryShea:
Commit message needs to say "Closes #886" or "Fixes #886" etc

JerryShea (approved):
approved - thanks - but please change commit message

Issue #894, opened 2024-05-16
Wire MethodReader fails to clear List before populating


alamar:
Please see the PR - if there is a fix please backport to x.24

alamar:
```
    // Definition of a DTO with a list-containing field.
    public static class ListContainingDto {
        Object list;
    }
```
(see the PR)
Changing type to List will work around the issue. Adding `extends AbstractMarshallableCfg` will also would.

tgd:
Thank you for the detailed reproducer Ilya. This is now scheduled, I will send an update soon.

hft-team-city:
Released in [Chronicle-Wire-2.26ea1](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.26ea1), [BOM-2.26ea10](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.26ea10)

hft-team-city:
Released in [Chronicle-Wire-2.25.21](https://github.com/OpenHFT/Chronicle-Wire/releases/chronicle-wire-2.25.21), [BOM-2.25.75](https://github.com/OpenHFT/OpenHFT/releases/chronicle-bom-2.25.75)

PR #900, opened 2024-06-05
Handle recycling of values in lists that are not in a List field, Fix…

peter-lawrey:
…es #894


alamar (approved):

src/main/java/net/openhft/chronicle/wire/SerializationStrategies.java
```diff
@@ -620,15 +620,22 @@ public Object readUsing(Class clazz, Object o, @NotNull ValueIn in, BracketType
             @NotNull List<Object> list = (o == null ? new ArrayList<>() : (List<Object>) o);
             @NotNull final WireIn wireIn = in.wireIn();
             long pos = wireIn.bytes().readPosition();
+            int count = 0;
             while (in.hasNextSequenceItem()) {
-                list.add(in.object());
-
+                if (list.size() > count) {
```
alamar, line 625: Perhaps we could switch CollectionFieldAccess to the same approach?

tgd (approved):

PR #903, opened 2024-06-10
Enable SerializableObjectTest.

### LINKED ISSUE #482: SerializableObjectTest is flappy
This could be a genuine bug. The error is:
```
org.junit.ComparisonFailure: expected:<[2,[ 1, ]3]> but was:<[2,[1,]3]>
Expected :[2, 1, 3]
Actual   :[2,1,3]
```
When you attach a debugger to it, the LHS (source) is a LinkedList containing integers 2, 1, 3 and the RHS (target) is an ArrayList with a single string element `2,1,3`

Oddly this behaviour seems non-deterministic, the test sometimes passes and sometimes fails.

yevgenp:
Closes #482

### DIFF
--- a/src/test/java/net/openhft/chronicle/wire/SerializableObjectTest.java
+++ b/src/test/java/net/openhft/chronicle/wire/SerializableObjectTest.java
@@ -20,11 +20,9 @@ package net.openhft.chronicle.wire;
 
 import io.github.classgraph.*;
 import net.openhft.chronicle.bytes.Bytes;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.DynamicTest;
 import org.junit.jupiter.api.TestFactory;
 
-import javax.sql.rowset.serial.SerialBlob;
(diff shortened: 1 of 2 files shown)


src/test/java/net/openhft/chronicle/wire/SerializableObjectTest.java
```diff
@@ -214,6 +213,7 @@ private static Stream<Object> reflectedObjects() {
             List<Object> objects = widgetClasses.stream()
                     .filter(c -> !IGNORED_CLASSES.contains(c.loadClass(true)))  // Filter out classes from the ignored list.
                     .filter(SerializableObjectTest::overridesEqualsObject)  // Ensure the class overrides equals() method.
+                    .filter(SerializableObjectTest::overridesToString)  // Ensure the class overrides equals() method.
```
tgd, line 216: Minor: Javadoc comment doesn't match filter

tgd (approved):

PR #909, opened 2024-06-21
Wire/issues/908

peter-lawrey:


src/main/java/net/openhft/chronicle/wire/JSONWire.java
```diff
@@ -373,6 +395,30 @@ public void copyOne(@NotNull WireOut wire, boolean inMap, boolean topLevel) thro
...
+                in.readPosition(position);
+                return false;
+            }
+        }
+        int ch = in.peekUnsignedByte();
+        if (Character.isLetterOrDigit(ch)) {
+            in.readPosition(position);
+            return false;
+        }
+        while (ch > 0 && ch <= ' ') {
```
tgd, line 414: Q: if this is going to return true by this point why does the iteration continue?

src/main/java/net/openhft/chronicle/wire/JSONWire.java
```diff
@@ -854,10 +900,17 @@ public JSONWire typeLiteral(@Nullable CharSequence type) {
         @Override
         public JSONValueOut typePrefix(@NotNull CharSequence typeName) {
             if (useTypes) {
-                startBlock('{');
+                boolean nested = bytes.peekUnsignedByte(bytes.writePosition() - 1) == '{';
```
tgd, line 903: Could `bytes.writePosition()-1` yield a value that breaks `peekUnsignedByte`?

tgd (approved):

src/main/java/net/openhft/chronicle/wire/JSONWire.java
peter-lawrey: ch = on line 416 takes the next character

src/main/java/net/openhft/chronicle/wire/JSONWire.java
peter-lawrey: peekUnsignedByte returns -1 if out of bounds.

PR #914, opened 2024-06-25
failing test for customer use case

JerryShea:


src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -2022,6 +2022,8 @@ else if (type == Set.class)
             componentType = extractClass(computeActualTypeArguments(Collection.class, field)[0]);
             if (componentType != Object.class) {
                 isLeaf = !Throwable.class.isAssignableFrom(componentType)
+                        // Don't recurse into the same class
+                        && !componentType.equals(field.getDeclaringClass())
```
david-ry4n, line 2026: Wasn't sure if this should be !componentType.isAssignmentFrom(field.getDeclaringClass())

JerryShea:
I created the PR so can't approve, but I do! Please wait for @peter-lawrey to give his blessing.

We will need to backport to 25 and 24 - hopefully that is not too difficult

And we should create an issue for release notes

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
JerryShea: I think equality is correct as the test whereby the types were not the same but one assignable from the other, seemed to work OK

david-ry4n:
This closes #915

peter-lawrey (approved):

PR #921, opened 2024-07-09
Added the the service name to the message history

RobAustin:
Tools like service- links don't work if two or more severs write to the same queue, because they are unable to determine from which services the message originated, this enriches the history message to include the service name.

see 

wire - https://github.com/OpenHFT/Chronicle-Wire/issues/922

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/MessageHistory.java
+++ b/src/main/java/net/openhft/chronicle/wire/MessageHistory.java
@@ -150,10 +150,10 @@ public interface MessageHistory extends Marshallable {
     boolean isDirty();
 
     /**
-     * @return the name of the service that last wrote the message history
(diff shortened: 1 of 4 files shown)


src/main/java/net/openhft/chronicle/wire/MessageHistory.java
```diff
@@ -148,4 +148,17 @@ default void doWriteHistory(DocumentContext dc) {
      * {@link Marshallable#writeMarshallable(net.openhft.chronicle.wire.WireOut)}
      */
     boolean isDirty();
+
+    /**
+     * @return the name of the service that last wrote the message history
+     */
+    default CharSequence serviceName() {
+        return null;
```
peter-lawrey: can you change this to empty string

george-ball (approved):

PR #927, opened 2024-07-19
Fix for Chronicle-Wire/issues/920

batespj:
Deserialisation using a provided object reference was not working as expected for JSONWire. Closer investigation showed that the default method defined in ValueIn was being called. The solution is to override the default method in JSONWire.

### DIFF
--- a/src/test/java/net/openhft/chronicle/wire/TestJsonIssue467.java
+++ b/src/test/java/net/openhft/chronicle/wire/TestJsonIssue467.java
@@ -22,7 +22,6 @@ import net.openhft.chronicle.bytes.Bytes;
 import net.openhft.chronicle.bytes.OnHeapBytes;
 import org.jetbrains.annotations.NotNull;
 import org.junit.Assert;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import static net.openhft.chronicle.core.pool.ClassAliasPool.CLASS_ALIASES;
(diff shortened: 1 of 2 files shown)


src/test/java/net/openhft/chronicle/wire/TestJsonIssue467.java
```diff
@@ -22,27 +22,26 @@
...
-import org.junit.Ignore;
 import org.junit.Test;
 
 import static net.openhft.chronicle.core.pool.ClassAliasPool.CLASS_ALIASES;
 
 // relates to https://github.com/OpenHFT/Chronicle-Wire/issues/467
 public class TestJsonIssue467 {
 
-    public static class ResponseItem extends SelfDescribingMarshallable {
+    public static class ResponseItem467 extends SelfDescribingMarshallable {
```
batespj, line 32: The tests pass without renaming this inner class, however tests then start to fail in JSONWithAMapTest. 
JSONWithAMapTest also defines an inner class called ResponseItem and the tests fail with the error:
"Did not replace by simpleName class net.openhft.chronicle.wire.TestJsonIssue467$ResponseItem with class net.openhft.chronicle.wire.marshallable.JSONWithAMapTest$ResponseItem"

I don't think it breaks these tests to give TestJsonIssue467's inner class ResponseItem the suffix '467' to avoid this collision of class names and this seems like the simplest solution.

tgd (approved):

JerryShea:
Please make sure your commit message has the correct form to ensure the issue is closed

tgd (approved):

PR #931, opened 2024-08-26
Ability to reuse multiple parameter objects in method reader

benbonavia:
Supports the use case where method reader parameters could be defined using abstract/parent objects which could have marshallable and non-marshallable implementations. Enable compliance with this scenario with `VanillaMethodReaderBuilder#multipleNonMarshallableParamTypes`

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
+++ b/src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
@@ -48,7 +48,7 @@ import static net.openhft.chronicle.wire.GenerateMethodWriter.isSynthetic;
  */
 public class GenerateMethodReader {
 
-     // Configuration flag for dumping the generated code.
+    // Configuration flag for dumping the generated code.
     private static final boolean DUMP_CODE = Jvm.getBoolean("dumpCode");
     // Set of interfaces that are not meant to be processed.
     private static final Set<Class<?>> IGNORED_INTERFACES = new LinkedHashSet<>();
@@ -78,7 +78,10 @@ public class GenerateMethodReader {
     // Configuration for the type of wire to use for serialization/deserialization.
     private final WireType wireType;
 
-     // Handlers for metadata during the method reader generation.
+    // Check for supporting parameters which can either be non-Marshallable or Marshallable
+    private final Boolean multipleNonMarshallableParamTypes;
+
+    // Handlers for metadata during the method reader generation.
     private final Object[] metaDataHandler;
 
     // Instances of the classes/interfaces for which method readers are to be generated.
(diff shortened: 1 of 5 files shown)


JerryShea (approved):
nice one

JerryShea:
@benbonavia is it possible for us to detect that the ` VanillaMethodReaderBuilder#multipleNonMarshallableParamTypes` has not been set but should be and perhaps throw a helpful exception?

peter-lawrey (changes requested):
Can you update the test to show how the outcome changes based on the value used.

peter-lawrey:
In particular if the value is not null, ie not the default, something needs to be added to the generated class name otherwise it won't change for subsequent calls. e.g. add a T or an F.

src/test/java/net/openhft/chronicle/wire/MethodWriterVagueTypesTest.java
```diff
@@ -67,65 +69,90 @@ public int hashCode() {
...
     private void test(Runnable methodCall, MethodReader reader, ArrayBlockingQueue<Object> queue, Object ... objs) throws Exception {
         methodCall.run();
+        boolean marshallableToNonMarshallable = false;
+        if (prevObjClasses[0] != null) {
+            for (int i=0; i < objs.length; i++) {
+                marshallableToNonMarshallable |= Marshallable.class.isAssignableFrom(prevObjClasses[i]) && !(objs[i] instanceof Marshallable);
+            }
+        }
+        if (Boolean.FALSE.equals(multipleNonMarshallableParamTypes) && marshallableToNonMarshallable) {
+            Assert.assertThrows(Exception.class, () -> assertWrite(reader, queue, objs));
```
JerryShea: cool

JerryShea: what exception does it throw?

JerryShea (approved):

src/test/java/net/openhft/chronicle/wire/MethodWriterVagueTypesTest.java
benbonavia: In this scenario, it either throws a RuntimeException from `assertWrite()` or an `IORuntimeException` is thrown by TextWire while marshalling - I can change it to a RuntimeException

src/main/java/net/openhft/chronicle/wire/GenerateMethodReader.java
```diff
@@ -933,6 +958,9 @@ private String generatedClassName0() {
             sb.append(wireType.toString()
                     .replace("_", ""));
 
+        // Append multi marshal/non-marshal support
+        sb.append(multipleNonMarshallableParamTypes != null ? Boolean.TRUE.equals(multipleNonMarshallableParamTypes) ? 'T' : 'F' : 'A');
```
JerryShea: The default should be to add nothing for backward compatability

src/main/java/net/openhft/chronicle/wire/VanillaMethodReaderBuilder.java
```diff
@@ -75,6 +75,9 @@ public class VanillaMethodReaderBuilder implements MethodReaderBuilder {
     // A flag to indicate whether the reader is in a scanning mode.
     private boolean scanning = false;
 
+    // A flag to determine support for parameters which can either be non-Marshallable or Marshallable
+    private Boolean multipleNonMarshallableParamTypes = false;
```
JerryShea: The default should be null but requires more fixes for all tests to pass.

peter-lawrey (approved):

PR #948, opened 2024-10-15
Remove/sdo write

peter-lawrey:


JerryShea (commented):
Could this change be made against develop?

tgd:
Looks that way - @peter-lawrey are you happy for this to land in develop first?

PR #965, opened 2024-11-14
Fixed incorrect generic return type detection in method writer

### LINKED ISSUE #963: Wire generates incorrect code when implementing Transport from Queue Web Gateway
When interface extends `Transport<T>` it has to redefine `T destination(...)` method explicitly, or method writer's generated code will not compile, and will not work correctly.

Workaround is known but the root case has to be addressed

alamar:
closes #963


alamar:
(branch naming is incorrect)

PR #967, opened 2024-11-14
Fix edge case when non-generic interface extends a generic one

### LINKED ISSUE #963: Wire generates incorrect code when implementing Transport from Queue Web Gateway

alamar:
closes #963


alamar:
depends on OpenHFT/Chronicle-Core#677

alamar:
https://teamcity.chronicle.software/buildConfiguration/Chronicle_BuildAll_Java21/1070963?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false
https://teamcity.chronicle.software/buildConfiguration/Chronicle_BuildAll_Release/1070962?hideTestsFromDependencies=false&hideProblemsFromDependencies=false
Tests pass

maxim-ponomarev (approved):

PR #975, opened 2024-11-25
Optimising the writing and reading of JSON

peter-lawrey:


microbenchmarks/src/main/java/net/openhft/chronicle/wire/benchmarks/ComparisonMain.java
```diff
@@ -120,27 +133,42 @@ public static void main(String... args) throws Exception {
...
+                    .measurementIterations(3)
+                    .threads(8)
+                    .forks(11)
+                    .mode(Mode.Throughput)
+                    .warmupTime(TimeValue.seconds(1))
+                    .measurementTime(TimeValue.seconds(time))
+                    .timeUnit(TimeUnit.SECONDS)
+                    .build();
+
+            new Runner(opt).run();
```
JerryShea, line 167: why fork?

microbenchmarks/src/main/java/net/openhft/chronicle/wire/benchmarks/ComparisonMain.java
```diff
@@ -120,27 +133,42 @@ public static void main(String... args) throws Exception {
...
         }
     }
 
-    @Benchmark
+    //    @Benchmark
```
JerryShea, line 171: why remove these attrs?

peter-lawrey:
This is a work in progress for a benchmark which isn't needed any more

PR #976, opened 2024-12-09
Fix JSON Wire null double breaking parse

### LINKED ISSUE #953: JSONWire parser is confused when reading a field after null resolves to Double NaN
df307774 defines that NaN is turned into `null` on writing, and `null` is turned into NaN double on reading.
However, it seems that parsing such `null` sets parser in some bad state where it can no longer parse next fields in the same object.

I will attach a reproducer test.

alamar:
Fix general case of double parsing not skipping commas
closes #953

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/TextWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/TextWire.java
@@ -1932,7 +1932,7 @@ public class TextWire extends YamlWireOut<TextWire> {
         @Override
         public <T> WireIn int32(@NotNull T t, @NotNull ObjIntConsumer<T> ti) {
             consumePadding();
-            ti.accept(t, (int) getALong());
+            ti.accept(t, (int) int64());
             return TextWire.this;
         }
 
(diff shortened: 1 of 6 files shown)


alamar:
https://teamcity.chronicle.software/buildConfiguration/Chronicle_BuildAll_BumpedBranch/1082115?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildChangesSection=true all tests pass
@tgd please review

src/main/java/net/openhft/chronicle/wire/TextWire.java
```diff
@@ -1932,7 +1932,7 @@ public <T> WireIn uint16(@NotNull T t, @NotNull ObjIntConsumer<T> ti) {
         @Override
         public <T> WireIn int32(@NotNull T t, @NotNull ObjIntConsumer<T> ti) {
             consumePadding();
-            ti.accept(t, (int) getALong());
+            ti.accept(t, (int) int64());
```
tgd, line 1935: Why does this accept int64 when the calling method is called int32?

(I see this is not a new issue, it used to be a long before)

src/main/java/net/openhft/chronicle/wire/TextWire.java
```diff
@@ -2743,6 +2734,7 @@ public long int64() {
 
             long l = getALong();
             checkRewind();
+            consumePadding(1);
```
tgd, line 2737: What is the significance of this call to `consumePadding(1)`?

src/main/java/net/openhft/chronicle/wire/TextWire.java
alamar: I really have no answer but I can add "enforce or throw" check here. I'm not advising to backport this change anyway and it seems safe to do in `ea`. WDYT?

src/main/java/net/openhft/chronicle/wire/TextWire.java
alamar: That's the whole point of this change. When a comma supercedes a number, it gets consumed for free, but if there's whitespace before a comma (permitted by all specs) we need to explicitly consume this single comma. It is really "up to one".

tgd (approved):

PR #983, opened 2025-01-01
JsonWire: Improve handling of double special values fixes #982

### LINKED ISSUE #982: JsonWire - Improve serialisation of Double NaN and Infinity
Historically JsonWire in some cases would emit 'null' for `Double.NaN' and the unquoted strings `Infinity` for `Double.Infinite`. In the latter case this token was not represented as a JSON string literal in double quotes so would create parse errors when that JSON was re-parsed down the line.

This ticket does the following:

- Maps Double.NaN, Double.POSITIVE_INFINITY, Double.NEGATIVE_INFINITY to the JSON strings `"NaN", "Infinity", "-Infinity"`
- Implements the same logic for `float`

tgd:
Includes fixes for Double.NaN, Double.POSITIVE_INFINITY, Double.NEGATIVE_INFINITY.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -743,6 +743,26 @@ class Data implements Marshallable {
 }
 ----
 
+== Wire type implementation details
+
+=== JSONWire
+
+==== Handling of special floating point values such as Double.NaN
+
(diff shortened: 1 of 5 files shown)


JerryShea (approved):
very nice clean PR

benbonavia (approved):

PR #990, opened 2025-02-24
941 Bugfix - JsonWire incorrectly decodes null collections

KealanTolandChronicle:
Update JSON wire handling to correctly decode null collections from string.
Refactor reproducer into unit test.
Remove unused import.


KealanTolandChronicle:
Benchmarked performance using async profiler with json decode test ran 10mil times across 3 potential solutions  - 

1. Indexed for loop (26586 samples)
2. Enhanced for loop (26718 samples)
3. Using util function WireMarshaller.getAllField with LinkedHashMap (30395 samples)

Looking at the allocation flame-graphs produced, seems the getAllField option is the slowest and has most memory overhead, potentially most useful if we require inherited fields too, however as the json string input will declare all fields necessary I have decided not to use this here.

The difference between the other two loops is very little, I'm guessing most modern JVMs can optimise these quite well but just to be on the safe side, I have opted for an indexed loop which should support high performance even on more obscure JVMs and introduces very little developer maintenance overhead for this simple change.

james-mcsherry (approved):

tgd (approved):

PR #996, opened 2025-05-30
implement missing ValueIn#sequence override for Text&YamlWire

JerryShea:
for Yaml & TextWire 
```
    <T> boolean sequence(List<T> list, @NotNull List<T> buffer, Supplier<T> bufferAdd, Reader reader0) throws InvalidMarshallableException;
```
was not properly implemented - it's implementation instead uses the overload without the `Reader`. 

This PR removes the stubbed implementations and ensures that the `Reader` argument is used. It also removes the  `net.openhft.chronicle.wire.ValueIn#sequence(java.util.List<T>, java.util.List<T>, java.util.function.Supplier<T>)` overloads from the Text Wires as the default impl. interface already implements the correct behaviour.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/TextWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/TextWire.java
@@ -2186,23 +2186,18 @@ public class TextWire extends YamlWireOut<TextWire> {
         }
 
         /**
-         * Handles the processing of a sequence, delegating to an overloaded version of itself.
+         * Handles the processing of a sequence.
          *
          * @param <T> The type of items in the lists.
          * @param list The main list that should be populated based on the buffer.
          * @param buffer A temporary buffer used for staging data.
          * @param bufferAdd A supplier function that can add items to the buffer.
-         * @param reader0 This seems to be an unused reader, possibly for future extensions.
+         * @param reader0 The reader that processes each item in the sequence.
          * @return Returns a boolean indicating success/failure or some other status.
          * @throws InvalidMarshallableException if there's an error during the sequence processing.
          */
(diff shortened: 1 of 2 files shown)


src/main/java/net/openhft/chronicle/wire/YamlWire.java
```diff
@@ -1779,12 +1774,7 @@ public <T> boolean sequence(@NotNull List<T> list, @NotNull List<T> buffer, @Not
                 int minIndent = yt.secondTopContext().indent;
                 yt.next(Integer.MAX_VALUE);
 
-                while (hasNextSequenceItem()) {
-                    if (buffer.size() <= list.size())
-                        buffer.add(bufferAdd.get());
-                    Object using = buffer.get(list.size());
-                    list.add((T) valueIn.object(using, using.getClass()));
-                }
+                reader0.accept(this, list, buffer, bufferAdd);
```
JerryShea, line 1777: Another dedupe, but this code did not call `if (t instanceof Resettable) ((Resettable) t).reset();` before, unlike other implementations in other Wires, so making this consistent is a nice by-product

src/main/java/net/openhft/chronicle/wire/TextWire.java
```diff
@@ -2224,14 +2219,7 @@ public <T> boolean sequence(@NotNull List<T> list, @NotNull List<T> buffer, @Not
...
 
-            while (hasNextSequenceItem()) {
-                int size = list.size();
-                if (buffer.size() <= size) buffer.add(bufferAdd.get());
-
-                final T t = buffer.get(size);
-                if (t instanceof Resettable) ((Resettable) t).reset();
-                list.add(object(t, (Class<T>) t.getClass()));
-            }
+            reader0.accept(this, list, buffer, bufferAdd);
```
JerryShea, line 2222: `ValueIn#reader0` already implements this, so this dedupes

PR #1204, opened 2025-06-26
Adv/01 vers2

peter-lawrey:
Update license and some docs


keiran-lawrey (commented):
Notes before approval.
- Graphs work thoughg the colour scheme should be improved
EventsByMethods.adoc Line:
- 34 - Unsure if the example `void onEvent(String eventType, WireIn data)` is accurate
- 97 - Serialisation is spelt with british english but maybe we don't want that
- 100 - "above half the latency" is changed to "about half the latency" 
this change in meaning may want reversing
- 104 - Appears useful but checking that these explanations are accurate is worth it
- 135 - StringBuilder field is changed to String which is probably wrong

PR #1208, opened 2025-07-22
Add class context and configurable warn/debug level for unknown-method logging

peter-lawrey:
**Motivation**
Today, when a method call can’t be resolved, the builder logs only the method name (or ID) and source history—which makes it hard to trace *which* target class was missing the implementation. It also always defaults to debug‐level logging, giving no easy way to promote warnings in production. By including the invoked class in the log message and driving warn vs. debug from a system property, users gain immediate visibility into missing handlers and can opt in to stricter logging when they need it.

**What’s Changed**

1. **`exceptionHandlerOnUnknownMethod` now driven by `chronicle.methodReader.warn`**

   * Defaults to `Jvm.warn()` when the property is true (else `Jvm.debug()`)
2. **`createDefaultParselet` extended to take an optional `Class<?> clazz`**

   * Overload marked `@Deprecated` for backward compatibility and delegates to the new two-arg form
3. **`errorMsg(...)` enhanced**

   * Prepends `"called on <class>"` (auto-unwrapping proxies and lambdas to their interface) when `clazz` is non-null
   * Switched to a `StringBuilder` for more efficient concatenation
4. **Builder `build(...)` now passes the first implementation’s class** (or `null` if none) into `createDefaultParselet`
5. **Tests updated** to supply an explicit `exceptionHandlerOnUnknownMethod(Jvm.debug())` and centralize their `ignoreException(...)` patterns

**Backward Compatibility**

* The one-arg `createDefaultParselet(ExceptionHandler)` remains available (deprecated) through  x.28–x.29 before removal.
* Default behavior remains `debug` if no property is set.

**Configuration**
Add to your `system.properties` (or pass via `-D`):

```properties
# When true, unknown-methods will be logged at WARN; otherwise DEBUG
chronicle.methodReader.warn=true
```


src/main/java/net/openhft/chronicle/wire/VanillaMethodReaderBuilder.java
```diff
@@ -106,29 +120,38 @@ public static WireParselet createDefaultParselet(ExceptionHandler exceptionHandl
...
      * @return A formatted error message for unrecognized methods or method IDs.
      */
     @NotNull
-    private static String errorMsg(CharSequence s, MessageHistory history, long sourceIndex) {
+    private static String errorMsg(CharSequence s, MessageHistory history, long sourceIndex, @Nullable Class<?> clazz) {
 
         // Determine whether the provided sequence is a method name or a method ID based on its first character.
         final String identifierType = s.length() != 0 && Character.isDigit(s.charAt(0)) ? "@MethodId" : "method-name";
-        String msg = "Unknown " + identifierType + "='" + s + "'";
+        StringBuilder msg = new StringBuilder()
```
tgd, line 142: Nit: This allocates a StringBuilder each time

tgd (approved):

src/main/java/net/openhft/chronicle/wire/VanillaMethodReaderBuilder.java
peter-lawrey: This path shouldn't be called much and is an improvement on the previous code

PR #1212, opened 2025-08-15
Update object instantiation to include Demarshallable types

james-mcsherry:
Updated objectMap so that a new instance is created if the target class implements Demarshallable or if the provided using instance is null, ensuring proper reconstruction of Demarshallable objects during deserialization.

Based off a support ticket, and has been tested against all existing tests with no issues.

JIRA: CORE-46


tgd (changes requested):
Hi @james-mcsherry please can you add a test that fails before the change, and passes after and link to jira. Thanks

tgd (approved):

PR #1237, opened 2025-11-18
Fix flaky test in AbstractMarshallableCfgTest by removing dependence on field…

lbh930:
### Description
Modified `AbstractMarshallableCfgTest.java` to sort the serialized fields before comparing them with the expected output. This ensures that the test is deterministic and doesn't depend on the iteration order of `Class.getDeclaredFields()`, which does not guarantee consistency.

### Motivation
NonDex detected test flakiness in `net.openhft.chronicle.wire.AbstractMarshallableCfgTest.asString` and the cause was that the test asserted a specific order of fields in the serialized YAML string. Since `SelfDescribingMarshallable` uses reflection to discover fields the order is not guaranteed. An example of NonDex output with detected failure is attached.
[nondex_output_AbstractMarshallableCfgTest.txt](https://github.com/user-attachments/files/23890151/nondex_output_AbstractMarshallableCfgTest.txt)


lbh930:
This PR is updated to make changes minimal.

PR #1238, opened 2025-11-18
Fix flaky test in Base64Test by removing dependence on field ordering

lbh930:
### Description
Modified `Base64Test.java` to sort the serialized fields before comparing them with the expected output. This ensures that the test is deterministic and doesn't depend on the iteration order of `Class.getDeclaredFields()`, which does not guarantee consistency.

### Motivation
NonDex detected test flakiness in `net.openhft.chronicle.wire.converter.Base64Test.onAnField` and the cause was that the test asserted a specific order of fields in the serialized YAML string. Since `SelfDescribingMarshallable` uses reflection to discover fields, the order is not guaranteed. An example of NonDex output with detected test failure is attached.
[nondex_output_base64test.txt](https://github.com/user-attachments/files/23890002/nondex_output_base64test.txt)


lbh930:
This PR is updated to make changes shorter.

PR #1245, opened 2026-01-20
Add newline-delimited `WireType.JSONL` (JSON Lines) alongside comma-separated `WireType.JSON_ONLY`

peter-lawrey:
This PR introduces **newline-separated JSON Lines support** via `WireType.JSONL`, while keeping the existing **comma-separated** stream behaviour available via `WireType.JSON_ONLY`.

`WireType.JSONL` is intended for streaming/log-style use cases where **each message is a complete JSON object on its own line** (`{...}\n{...}\n`). It also adds a JSON-specific reader that can safely consume both **JSONL** and legacy **comma-separated JSON streams**.

#### Motivation

`WireType.JSON_ONLY` currently emits a comma-separated JSON stream, which is awkward for line-oriented tooling and streaming pipelines. JSONL is a widely used convention that makes record boundaries explicit, supports incremental processing, and plays nicely with “tail -f”, log shippers, and batch ingestion.

#### What changed

* **New `WireType.JSONL`**

  * Text wire that outputs **one JSON document per line** (newline terminated).
  * Defaults to **no type metadata** (`useTypes(false)`), and uses JSONL-friendly conventions.

* **`JSONWire` JSONL mode (`useJsonlDocuments`)**

  * Adds a `jsonlMode` flag to preserve JSONL contexts.
  * Configures JSONL defaults:

    * `useTypes(false)`
    * `useLowerCaseEnums(true)` (configurable)
    * JSONL-specific read/write contexts
  * Overrides `methodWriter()` in JSONL mode to avoid the parent `useTextDocuments()` call overwriting the JSONL contexts.

* **`JSONLReadDocumentContext`**

  * JSON-specific read context that finds document boundaries by **tracking brace/array depth** (and respecting JSON strings/escapes), rather than YAML markers.
  * Accepts both:

    * **newline-delimited JSONL** (`{...}\n{...}\n`)
    * **comma-separated JSON streams** (`{...},{...},{...}`)
  * Strips outer braces when presenting the document body to the reader (matching the existing “document wrapper” behaviour).

* **`JSONLWriteDocumentContext`**

  * Ensures each document is a complete JSON object:

    * Wraps output in `{ ... }`
    * Appends `\n` after each document
  * Produces valid JSONL output for method writer streaming:

    * `{"event":[...]}`
    * `{"event":[...]}`

* **Lowercase enum serialisation**

  * Adds `useLowerCaseEnums(boolean)` and implements lowercase output for enums (including untyped enum values) without allocating new strings.

#### Output format examples

* **Method writer (JSONL)**

  ```
  {"event":["first",1 ]}
  {"event":["second",2 ]}
  {"event":["third",3 ]}
  ```

* **Object write note**

  * Double braces are expected in some paths because the **document context adds outer braces** and the **reader strips them** when exposing content.

#### Test plan

* Added/expanded `JSONLTest` with streaming coverage:

  * `testMethodReaderStreamingJsonl` – reads multiple newline-delimited events
  * `testMethodReaderCommaSeparatedJson` – reads comma-separated JSON streams
  * `testWriteJsonReadJsonl` – verifies `WireType.JSON_ONLY` output is readable by the JSONL reader
  * `testWriteJsonlReadJson` – verifies JSONL round-trip
  * Existing JSONL suite: **33 tests pass (1 skipped)**

    * `mvn -Dtest=JSONLTest test`

#### Compatibility

* `WireType.JSON_ONLY` behaviour remains unchanged.
* `WireType.JSONL` provides a line-oriented alternative and a reader tolerant of both newline and comma delimiters, easing migration and mixed-stream handling.


src/main/java/net/openhft/chronicle/wire/JSONWire.java
```diff
@@ -763,6 +792,179 @@ public void close() {
...
+                }
+                break;
+            }
+            return p;
+        }
+
+        @Override
+        public void close() {
+            super.close();
+        }
```
tgd: **Suggestion:** This empty `close()` override just calls `super.close()` and can be removed - it adds no behavior.

src/main/java/net/openhft/chronicle/wire/JSONWire.java
```diff
@@ -29,18 +29,21 @@
...
- * Represents the JSON wire format.
+ * Represents the JSON wire format for text-based marshalling and parsing.
  * <p>
  * This class provides functionality for managing JSON data in a wire format.
  * It currently provides a subset of functionalities similar to the YAML wire format.
  * The core capability of this class is to handle JSON data structures as {@code Bytes}
  * objects, allowing for efficient manipulation and parsing.
  */
-@SuppressWarnings("this-escape")
+@SuppressWarnings({"this-escape", "deprecation"})
```
tgd: **Suggestion:** Consider localizing the `@SuppressWarnings("deprecation")` to the specific method(s) that use deprecated APIs rather than suppressing it class-wide. This makes it clearer which code paths touch deprecated APIs.

src/main/java/net/openhft/chronicle/wire/JSONWire.java
```diff
@@ -1303,6 +1555,81 @@ public boolean useTypes() {
...
 
+    /**
+     * Configures this wire for JSONL (JSON Lines) output format.
+     * JSONL writes one JSON object per line with no type prefixes.
+     * Each document is terminated by a newline character.
+     *
+     * @return this wire configured for JSONL
+     * @see <a href="https://jsonlines.org/">JSON Lines specification</a>
+     */
+    public JSONWire useJsonlDocuments() {
```
tgd, line 1557: **Suggestion:** Consider adding a thread-safety note to the Javadoc, e.g.:
```java
 * <p>Thread safety: The returned wire instance is not thread-safe.
```
This matches the documentation style in the test class.

tgd (approved):
Well-implemented JSONL support with excellent test coverage. Approving with minor suggestions.

src/main/java/net/openhft/chronicle/wire/JSONWire.java
peter-lawrey: Good find

src/main/java/net/openhft/chronicle/wire/JSONWire.java
peter-lawrey: Good find

src/main/java/net/openhft/chronicle/wire/JSONWire.java
peter-lawrey: All of Wire is single threaded, not worth adding here in particular

PR #1246, opened 2026-01-27
CORE-49 fix nested lists with single entry parsing

benbonavia:
TextWire supports single item arrays without square bracket `[` placement. However, if there is a list of multiple items and one of the items has a single array within it without square brackets, all subsequent items in the parent list are not read with an error thrown.

The issue here is that within `TextWire$TextValueIn` there is a variable called `sequenceLimit` which keeps track of how many items should be in a list - this is primarily to handle the case of a singleton list. This breaks when there is an outer list with multiple elements and the inner singleton list is not the last element.

The solution here is to remove this field and make it a local variable and add a new `savedValue` into the parse stack to keep track of the limit for the current marshallable read. This way the outer list limit value is not clobbered by the inner list limit value.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
     <parent>
         <groupId>net.openhft</groupId>
         <artifactId>java-parent-pom</artifactId>
-        <version>2026.XYZ</version>
+        <version>1.27ea1</version>
         <relativePath/>
     </parent>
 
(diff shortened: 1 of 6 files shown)


james-mcsherry (approved):
Looks good @benbonavia - one of the poms just has a conflict then good to go.

PR #1256, opened 2026-03-03
CORE-53: Fix StringBuilder[] deserialization producing array with identical elements

james-mcsherry:
## Summary

  - Fix `StringBuilder[]` deserialization where all array elements were the same pooled `StringBuilder` instance, causing every element to contain the last-written value
  - The `ScalarStrategy` for `StringBuilder` previously used `Wires.acquireStringBuilderScoped()` to obtain a pooled instance when `using` was `null` — during array deserialization each element is read with `using=null`, so the pool returned the same object for
   every slot
  - Replaced the pooled allocation with `new StringBuilder()`, matching the pattern used by other scalar strategies (String, Date, UUID, etc.)

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/Wires.java
+++ b/src/main/java/net/openhft/chronicle/wire/Wires.java
@@ -1523,13 +1523,11 @@ public enum Wires {
 
                 case "java.lang.StringBuilder":
                     return ScalarStrategy.of(StringBuilder.class, (o, in) -> {
-                        StringBuilder builder;
-                        try (ScopedResource<StringBuilder> stlSb = Wires.acquireStringBuilderScoped()) {
-                             builder = (o == null)
-                                    ? stlSb.get()
-                                    : o;
-                            in.textTo(builder);
-                        }
+                        StringBuilder builder = (o == null)
+                                ? new StringBuilder()
+                                : o;
+                        if (in.textTo(builder) == null)
+                            return null;
                         return builder;
                     });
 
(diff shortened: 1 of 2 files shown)


KealanTolandChronicle (approved):
Looks great, solid fix and good test coverage.

src/main/java/net/openhft/chronicle/wire/Wires.java
```diff
@@ -1523,13 +1523,11 @@ public SerializationStrategy apply(@NotNull Class<?> aClass) {
...
                     return ScalarStrategy.of(StringBuilder.class, (o, in) -> {
-                        StringBuilder builder;
-                        try (ScopedResource<StringBuilder> stlSb = Wires.acquireStringBuilderScoped()) {
-                             builder = (o == null)
-                                    ? stlSb.get()
-                                    : o;
-                            in.textTo(builder);
-                        }
+                        StringBuilder builder = (o == null)
+                                ? new StringBuilder()
```
benbonavia, line 1527: This concerns me as it has potential to introduce additional garbage during deserialising. The purpose of the `ScopedResource` is to grab an existing StringBuilder from a pool (the size of which is controlled by `chronicle.stringBuilderPool.instancesPerThread` which defaults to 4 per thread). This reuse would be to not create excessive garbage. Having said that, the `ScalarStrategy` and read function will create garbage after processing the current `ValueIn`. 

Ultimately, while I feel the best solution would be to alter the scope for the resource - which would be deserialisation as a whole, this would not be immediately possible without considerable rework efforts and potential side affects.

benbonavia (approved):

PR #1259, opened 2026-04-28
Add `@FieldOrder` for explicit marshalling field order

peter-lawrey:
This PR adds a new `@FieldOrder` type annotation to Chronicle Wire so `Marshallable` classes can define a stable, explicit serialisation order for fields declared directly on the annotated class.

## Why

Some serialisation paths and tests currently depend on reflection-discovered field order, which can be brittle across JVMs, compiler output, or source changes. This is especially important for text/YAML assertions and `FIELDLESS_BINARY`, where field order is part of the compatibility contract.

## What changed

* Added `@FieldOrder(String[] value)` for declaring marshalling order at type level.
* Updated `WireMarshaller.getAllField(...)` to honour `@FieldOrder` per class in an inheritance hierarchy.
* Added validation for:

  * missing marshallable fields
  * duplicate names, including after trimming
  * blank names
  * unknown fields
  * static/transient/non-marshallable fields
  * inherited fields listed on the wrong class
* Preserved existing default reflection-discovery behaviour when `@FieldOrder` is absent.
* Added tests covering inheritance, stable `SelfDescribingMarshallable` output, validation failures, whitespace trimming, and `FIELDLESS_BINARY` order mismatch behaviour.
* Applied `@FieldOrder` to affected test DTOs and examples to stabilise expected output.
* Updated Wire documentation and project requirements to describe `@FieldOrder` and clarify its role in schema evolution and fieldless binary formats.

## Notes

`@FieldOrder` applies only to fields declared directly on the annotated class. Inherited fields keep the order defined by their own declaring class. For `FIELDLESS_BINARY`, this can stabilise the marshaller order, but reader and writer must still agree on field order, field types, and field count.

### DIFF
--- a/src/main/docs/project-requirements.adoc
+++ b/src/main/docs/project-requirements.adoc
@@ -74,7 +74,7 @@ Schema Evolution and Versioning ::
 * **Comparison:** Explicit comparison of Wire's schema evolution capabilities with those of Protobuf and Avro.
 
 Annotations for Fine-Grained Control and Customisation ::
-* For each key annotation (e.g., `@LongConversion`, `@NanoTime`, `@Base64`, `@Base85`, `@ShortText`, `@MaxUtf8Len`, `@Comment`, `@FieldNumber`):
+* For each key annotation (e.g., `@LongConversion`, `@NanoTime`, `@Base64`, `@Base85`, `@ShortText`, `@MaxUtf8Len`, `@Comment`, `@FieldNumber`, `@FieldOrder`):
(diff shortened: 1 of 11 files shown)


tgd (changes requested):
Please can you link a jira with the release notes field filled to describe the change

peter-lawrey:
The change now has a Jira with Release Notes https://chroniclesoftware.atlassian.net/browse/CORE-59

tgd (approved):

PR #1262, opened 2026-06-01
CORE-65: Write faithful finite JSON float and double numbers

peter-lawrey:
## Context

Part of CORE-65: make finite float/double values round-trip faithfully through Chronicle's text wires. The bug split across two layers: the writer (small-double formatting in `YamlWireOut`) and the reader (`Maths.asDouble` not correctly rounded on common decimal inputs).

Stacked PRs:
1. `OpenHFT/Chronicle-Core` — `Maths.asDouble` Clinger fast path, regression tests, and accuracy docs.
2. `OpenHFT/OpenHFT` / `chronicle-bom` — point consumers at snapshots carrying the CORE-65 stack.
3. `OpenHFT/Chronicle-Bytes` — `parseDouble` precision regression suite and accuracy docs.
4. `OpenHFT/Chronicle-Wire` — write all finite doubles/floats as faithful JSON numbers and cover formatter thresholds.

Merge/deploy order: Core first, then BOM, then Bytes and Wire. This branch is stacked on the CORE-64 JSON finite-number fix, so CORE-64 should merge first or this branch should be rebased after it lands.

## What Changed

`YamlWireOut.float64` now routes values below `1e-3` through `Double.toString` instead of the old compacting path:

- Previous lower fast-path band included `[1e-7, 1e-3)`.
- That band used `Math.round(d * 1e16) / 1e9`, which could truncate high-precision small magnitudes.
- The lower fast-path bound is now `1e-3`, so sub-`1e-3` finite values use the faithful `Double.toString` path.

For JSON output, finite floats/doubles of every magnitude are written as unquoted JSON numbers. Non-finite values (`NaN`, infinities) remain quoted JSON string literals because JSON has no numeric representation for them.

The acceptance test now checks writer faithfulness independently from Chronicle's own reader: the emitted JSON numeric token, parsed by the JDK reference parser, must denote the exact input value.

## Why

The writer should never emit a finite JSON number token that denotes a different value from the input. Clean values remain byte-identical to the old output; the changed outputs are the previously truncated high-precision small values.

Measured writer sweep data for this bug class showed 0 unfaithful writes across about 13M finite doubles below `1e15` after this change.

## Risk

`float64` lives in shared `YamlWireOut`, so this can also affect YAML/Text formatting for small finite values in `[1e-7, 1e-3)`. This is intended for faithfulness, but reviewers with cross-language consumers or stored golden files of small numbers should inspect that output change explicitly.

The reader-side exactness for all realistic-path values depends on the CORE-65 Core/BOM/Bytes stack. This PR's writer faithfulness is checked independently from Chronicle's reader, and end-to-end reader checks are scoped to values the current parser restores exactly.

## Verification

- `mvn -q -Dtest=JsonWireDoubleAndFloatSpecialValuesAcceptanceTests test`

Full CI should cover the wider JSON/YAML/Text golden-master surface before merge.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/YamlWireOut.java
+++ b/src/main/java/net/openhft/chronicle/wire/YamlWireOut.java
@@ -1004,28 +1004,19 @@ public abstract class YamlWireOut<T extends YamlWireOut<T>> extends AbstractWire
             double ad = Math.abs(d);
             if (ad == 0) {
                 bytes.append(d);
-            } else if (ad >= 1e-7 && ad < 1e15) {
+            } else if (ad >= 1e-3 && ad < 1e15) {
                 if ((int) (ad / 1e6) * 1e6 == ad) {
                     bytes.append((int) (d / 1e6)).append("E6");
                 } else if ((int) (ad / 1e3) * 1e3 == ad) {
                     bytes.append((int) (d / 1e3)).append("E3");
-                } else if (ad < 1e-3) {
-                    double d7 = Math.round(d * 1e16) / 1e9;
-                    double ad7 = Math.abs(d7);
-                    if (ad7 < 1e1)
-                        bytes.append(d7).append("E-7");
-                    else if (ad7 < 1e2)
-                        bytes.append(d7 / 1e1).append("E-6");
-                    else if (ad7 < 1e3)
-                        bytes.append(d7 / 1e2).append("E-5");
-                    else if (ad7 < 1e4)
-                        bytes.append(d7 / 1e3).append("E-4");
-                    else
-                        bytes.append(d7 / 1e4).append("E-3");
                 } else {
                     bytes.append(d);
                 }
             } else {
+                // |d| < 1e-3 (including the former [1e-7, 1e-3) band), |d| >= 1e15, and NaN/Infinity
+                // all delegate to writeSpecialDoubleValueToBytes, which formats finite values with the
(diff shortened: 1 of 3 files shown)


src/main/java/net/openhft/chronicle/wire/YamlWireOut.java
```diff
@@ -1004,28 +1004,19 @@ public T float64(double d) {
...
-                    else
-                        bytes.append(d7 / 1e4).append("E-3");
                 } else {
                     bytes.append(d);
                 }
             } else {
+                // |d| < 1e-3 (including the former [1e-7, 1e-3) band), |d| >= 1e15, and NaN/Infinity
+                // all delegate to writeSpecialDoubleValueToBytes, which formats finite values with the
+                // faithful Double.toString. The previous Math.round(d * 1e16) / 1e9 compaction over
+                // [1e-7, 1e-3) silently truncated high-precision small magnitudes (CORE-64).
```
tgd, line 1019: Hi @peter-lawrey - if I am reading this right this will create more garbage now at runtime for these values. What are your thoughts on this?

tgd (changes requested):

src/main/java/net/openhft/chronicle/wire/YamlWireOut.java
peter-lawrey: There is more garbage, but greater accuracy. I am assuming values less than 0.001 are rare; we can revisit this if it turns out a client needs them to be more efficient.

tgd (approved):

PR #1264, opened 2026-06-12
Bugfix/bytes length32Fix unsigned BYTES_LENGTH32 handling and guard nested BinaryWire read limits

peter-lawrey:
This change tightens `BinaryWire` length-prefixed reads so `BYTES_LENGTH32` values are handled as unsigned 32-bit lengths and cannot extend the current read limit.

It adds a `long`-based `readWithLength` path, validates that supplied lengths fit the unsigned 32-bit range, and rejects malformed or unchecked input before it can corrupt the nested read limit.

### Problem

`BYTES_LENGTH32` was previously read using signed `readInt()` in some paths.

For encoded lengths with the top bit set, this could produce a negative `int` length. That made the nested limit calculation unsafe, because the existing check only rejected limits greater than the current limit and did not explicitly reject negative or oversized unsigned lengths.

In malformed or unchecked input, this could allow parsing to continue with a corrupt read limit rather than failing at the length boundary.

### Changes

* Read `BYTES_LENGTH32` with `readUnsignedInt()` instead of signed `readInt()`.
* Promote length handling from `int` to `long` where required.
* Add a `readWithLength(WireOut, long)` overload for unsigned 32-bit lengths.
* Keep the existing `readWithLength(WireOut, int)` method as a delegating compatibility overload.
* Add explicit validation for negative and greater-than-32-bit lengths.
* Add `guardedReadLimit(long)` to ensure nested values stay inside the enclosing read limit.
* Reuse the guarded limit calculation in `objectWithInferredType`.
* Improve failure messages when a declared length exceeds the bytes remaining before the current read limit.

### Tests

Adds coverage for:

* Copying a length-prefixed map fragment through the new `long` overload.
* Rejecting negative lengths.
* Rejecting lengths greater than unsigned 32-bit.
* Treating `BYTES_LENGTH32` values with the top bit set as unsigned.
* Rejecting `0xFFFFFFFF` length values before extending the read limit.
* Rejecting malformed unchecked object input before corrupting the read limit.

### Risk

Low to medium.

The change affects binary length-prefixed read paths, including malformed input handling. Valid encoded data should continue to read as before, while invalid or truncated data now fails earlier with a clearer `IORuntimeException`.

### Validation

Run the updated `BinaryWireReadWithLengthTest` test suite.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
@@ -485,14 +485,14 @@ public class BinaryWire extends AbstractWire implements Wire {
                     // Handle byte lengths and read accordingly.
                     case BYTES_LENGTH8: {
                         bytes.uncheckedReadSkipOne();
-                        int len = bytes.readUnsignedByte();
+                        long len = bytes.readUnsignedByte();
                         readWithLength(wire, len);
                         break outerSwitch;
                     }
 
                     case BYTES_LENGTH16: {
                         bytes.uncheckedReadSkipOne();
-                        int len = bytes.readUnsignedShort();
+                        long len = bytes.readUnsignedShort();
                         readWithLength(wire, len);
                         break outerSwitch;
                     }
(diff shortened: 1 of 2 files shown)


benbonavia:
This work has been completed from a separate, consolidated branch

PR #1266, opened 2026-06-18
CORE-68: Fix BinaryWire unsigned lengths and read-limit guards

sam-ross:
## Summary

This PR tightens `BinaryWire` handling of declared lengths and read boundaries.

It fixes unsigned `BYTES_LENGTH32` decoding, validates length-prefixed reads before narrowing read limits, and closes several malformed-input edge cases where byte reads could continue past the current value boundary or fail unclearly.

## Problem

`BYTES_LENGTH32` represents an unsigned 32-bit length, but some paths read it as a signed `int`. Encoded lengths with the top bit set could therefore be interpreted as negative values, which made read-limit checks unsafe.

There were also byte-reading edge cases around malformed or unchecked input:

  * nested length-prefixed values could claim more data than was available;
  * `PADDING32` text reads could proceed without enough bytes for the padding length and following code;
  * conversion reads, for example reading a non-byte-array value as bytes, could read beyond the declared value body;
  * zero-length byte-array reads could try to parse a value body that was not present;
  * reused output buffers could retain stale bytes when reading default byte values.

## Changes

  * Read `BYTES_LENGTH32` with `readUnsignedInt()`.
  * Promote length handling to `long` where unsigned 32-bit values must be represented.
  * Add `readWithLength(WireOut, long)` while keeping the existing `int` overload as a compatibility delegate.
  * Reject negative lengths and lengths greater than the unsigned 32-bit range.
  * Add `guardedReadLimit(long)` to keep nested length-prefixed reads inside the current read limit.
  * Apply guarded limits when reading nested length-prefixed object content.
  * Detect truncated `PADDING32` text reads before consuming padding length and the following value code.
  * Bound non-byte-array to bytes conversion to the declared value length.
  * Return an empty byte array for zero-length byte-array reads.
  * Clear `DefaultValueIn.bytes(BytesOut)` output before writing a default value.

## Tests

Adds coverage for:

  * Copying a length-prefixed fragment through the new `long` overload.
  * Rejecting negative and greater-than-unsigned-32-bit lengths.
  * Treating `BYTES_LENGTH32` values with the top bit set as unsigned.
  * Rejecting oversized declared lengths before extending the read limit.
  * Guarding malformed unchecked object and bytes reads.
  * Rejecting truncated `PADDING32` text reads.
  * Preventing conversion reads from consuming following wire data.
  * Returning an empty byte array for a zero-length byte-array body.
  * Preserving valid `BYTES_LENGTH32` byte-array reads.

## Risk

Low to medium.

This touches binary length-prefixed read paths and malformed input handling. Valid encoded data should continue to read as before. Invalid, truncated, or unchecked data should now fail earlier with clearer boundary checks, or return empty data where the declared value length is zero.

## Validation

Ran:

  * `mvn -q -Dtest=BinaryWireReadWithLengthTest test`

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
@@ -1959,6 +1959,7 @@ public class BinaryWire extends AbstractWire implements Wire {
                     case PADDING:
                         return readText(bytes.readUnsignedByte(), sb);
                     case PADDING32:
+                        // 4-byte padding length + 1-byte trailing value code
                         if (bytes.readRemaining() < 5)
                             throw new DecoratedBufferUnderflowException("Requires at least five bytes, readRemaining: " + bytes.readRemaining());
                         bytes.readSkip(bytes.readUnsignedInt());
--- a/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
@@ -485,14 +485,14 @@ public class BinaryWire extends AbstractWire implements Wire {
                     // Handle byte lengths and read accordingly.
                     case BYTES_LENGTH8: {
                         bytes.uncheckedReadSkipOne();
-                        int len = bytes.readUnsignedByte();
+                        long len = bytes.readUnsignedByte();
                         readWithLength(wire, len);
                         break outerSwitch;
                     }
 
                     case BYTES_LENGTH16: {
                         bytes.uncheckedReadSkipOne();
-                        int len = bytes.readUnsignedShort();
+                        long len = bytes.readUnsignedShort();
                         readWithLength(wire, len);
                         break outerSwitch;
                     }
@@ -671,17 +671,17 @@ public class BinaryWire extends AbstractWire implements Wire {
      * @param wire The wire output stream to write data to.
      * @param len  The length of data to be read, as an unsigned 32-bit value (0 to 4294967295).
      * @throws InvalidMarshallableException If there's an issue during marshalling.
-     * @throws IORuntimeException If the length is outside the unsigned 32-bit range.
+     * @throws IORuntimeException If the length is outside the unsigned 32-bit range,
+     *                            or exceeds the data remaining before the read limit.
      */
     @SuppressWarnings("incomplete-switch")
     public void readWithLength(@NotNull WireOut wire, long len) throws InvalidMarshallableException {
         // A length over 32-bit unsigned is unexpected: BYTES_LENGTH32 cannot encode it.
         if (len < 0 || len > 0xFFFF_FFFFL)
-            throw new IORuntimeException("Invalid length " + len + ", expected an unsigned 32-bit value");
+            throw new IORuntimeException("Invalid length " + len + ", expected an unsigned 32-bit value (0 to 4294967295)");
         long limit = bytes.readLimit();
         long newLimit = guardedReadLimit(len);
-        if (newLimit > limit)
-            throw new IORuntimeException("Can't extend the limit");
+
         try {
             bytes.readLimit(newLimit);
             @NotNull final ValueOut wireValueOut = wire.getValueOut();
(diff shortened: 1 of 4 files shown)


src/main/java/net/openhft/chronicle/wire/BinaryWire.java
```diff
@@ -485,22 +485,22 @@ private void copyOne(@NotNull WireOut wire, boolean first) throws InvalidMarshal
...
-                        int len = bytes.readUnsignedByte();
+                        long len = bytes.readUnsignedByte();
                         readWithLength(wire, len);
                         break outerSwitch;
                     }
 
                     case BYTES_LENGTH16: {
                         bytes.uncheckedReadSkipOne();
-                        int len = bytes.readUnsignedShort();
+                        long len = bytes.readUnsignedShort();
```
benbonavia, line 495: unsigned short and unsigned byte can safely represented inside an int

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
```diff
@@ -660,12 +660,28 @@ protected void fieldAnchor(@NotNull WireOut wire) {
...
+     * @param wire The wire output stream to write data to.
+     * @param len  The length of data to be read, as an unsigned 32-bit value (0 to 4294967295).
+     * @throws InvalidMarshallableException If there's an issue during marshalling.
+     * @throws IORuntimeException If the length is outside the unsigned 32-bit range,
+     *                            or exceeds the data remaining before the read limit.
+     */
+    @SuppressWarnings("incomplete-switch")
+    public void readWithLength(@NotNull WireOut wire, long len) throws InvalidMarshallableException {
+        // A length over 32-bit unsigned is unexpected: BYTES_LENGTH32 cannot encode it.
+        if (len < 0 || len > 0xFFFF_FFFFL)
```
benbonavia, line 680: the parameter `len` can not only be an unsigned 32-bit value but also an unsigned 8-bit or 16-bit value by its calls on lines 488 and 495 - perhaps a max value could be supplied?

src/main/java/net/openhft/chronicle/wire/DefaultValueIn.java
```diff
@@ -78,6 +78,7 @@ public Bytes<?> textTo(@NotNull Bytes<?> bytes) {
     @NotNull
     @Override
     public WireIn bytes(@NotNull BytesOut<?> toBytes) {
+        toBytes.clear();
```
benbonavia, line 81: is this required within the scope of these changes? I feel this could break some use cases

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
```diff
@@ -1919,6 +1959,8 @@ protected Bytes<?> writeCode(int code) {
                     case PADDING:
                         return readText(bytes.readUnsignedByte(), sb);
                     case PADDING32:
+                        if (bytes.readRemaining() < 5)
```
benbonavia, line 1963: why do we require at least 5 bytes?

benbonavia:
Please link Jira ticket in the PR title

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
james-mcsherry: As per @sam-ross and @peter-lawrey design doc - Intentional. They are widened to long so all three length cases call the single readWithLength(long) overload; the widening is implicit and free, and keeps the call sites uniform with the 32-bit path that genuinely needs long.

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
james-mcsherry: Detailed in design doc.   0xFFFFFFFF is the maximum BYTES_LENGTH32 can encode. The 8-bit and 16-bit callers are already bounded by readUnsignedByte/readUnsignedShort upstream.

src/main/java/net/openhft/chronicle/wire/DefaultValueIn.java
james-mcsherry: Called out explicitly in the design doc - Deliberate and in scope. Without it, toBytes could retain stale bytes from an earlier read when the default value is empty or shorter, leaking data that was never part of the default. This brings the path in line with ValueIn.bytes, which already clears.

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
james-mcsherry: Four bytes for the PADDING32 length plus one for the value code that follows. Detailed in doc.

james-mcsherry: Code comment has been added too.

james-mcsherry (approved):

PR #1269, opened 2026-07-30
CORE-75: Preserve a null Collection<String> when overwrite is false

james-mcsherry:
## CORE-75: Preserve a null Collection<String> when overwrite is false                                                                                                                                              
                                                                                                                                                                                                                      
  ### Issue                                                                                                                                                                                                         
  `StringCollectionFieldAccess.readValue` assigned a fresh collection before reading and only restored null when `overwrite` was true. A field serialised as `collection: !!null ""` therefore came back as an empty collection under `Wires.readMarshallable(this, wire, false)`, so a null `Set<String>`/`List<String>` did not survive a round trip. Only `String` component types were affected; `CollectionFieldAccess` and  `MapFieldAccess` already handled this correctly.                                                                                                                                                                    
                                                                                                                                                                                                                      
  ### Fix                                                                                                                                                                                                             
  Fall back to the `defaults` instance when the value is not a sequence, as `CollectionFieldAccess.readValue` already does. Consulting `defaults` rather than setting plain null keeps inline-initialised fields (`= new ArrayList<>()`) resolving to empty.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
+++ b/src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
@@ -2296,9 +2296,16 @@ public class WireMarshaller<T> {
             } else if (!coll.isEmpty()) {
                 coll.clear();
             }
-            boolean sequenced = read.sequence(coll, seqConsumer);
-            if (overwrite && !sequenced) {
-                field.set(o, null);
+            if (!read.sequence(coll, seqConsumer)) {
+                // A null in the input must survive; with overwrite there is no default to fall back to
+                Collection defaultColl = overwrite ? null : (Collection) field.get(defaults);
+                if (defaultColl == null) {
+                    field.set(o, null);
+                } else {
+                    coll.clear();
+                    if (!defaultColl.isEmpty())
+                        coll.addAll(defaultColl);
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


tgd (approved):

src/main/java/net/openhft/chronicle/wire/WireMarshaller.java
```diff
@@ -2296,9 +2296,16 @@ protected void readValue(Object o, Object defaults, ValueIn read, boolean overwr
             } else if (!coll.isEmpty()) {
                 coll.clear();
             }
-            boolean sequenced = read.sequence(coll, seqConsumer);
-            if (overwrite && !sequenced) {
-                field.set(o, null);
+            if (!read.sequence(coll, seqConsumer)) {
+                // A null in the input must survive; with overwrite there is no default to fall back to
+                Collection defaultColl = overwrite ? null : (Collection) field.get(defaults);
```
benbonavia, line 2301: this looks correct. Unfortunately other implementations of `readValue()` for Collection/Map field access are inconsistent - they should all have the same behaviour. Can you please update those?

benbonavia (changes requested):

james-mcsherry:
@benbonavia can you please re-review?

benbonavia (approved):

tgd (approved):

peter-lawrey (approved):

PR #1270, opened 2026-08-07
QUEUE-144: Add document context count contract

peter-lawrey:
QUEUE-143: Wire prerequisite for the Queue/CQE integration programme

## Summary

- Adds the `int DocumentContext.contextCount()` contract and propagates it through document context holders and wrappers.
- Advances writable Wire counts on `reset()` while `clear()` retains the current count; closed or unavailable contexts return `-1`.
- Adds `ProgressiveContext` with guidance for keeping the last count in a transient DTO field and writing context only when the count increases.

## Purpose and stack position

Context data must be resent when an output moves to a new context, such as after a Wire reset, a transport reconnect, or a Queue roll. Callers need to detect that change without depending on the concrete output implementation. The non-decreasing `contextCount()` contract provides that signal: an unchanged count means the existing context still applies, while an increased count means it must be sent again.

This contract is isolated in the first PR so its semantics and `int` representation can be reviewed independently of listener registration and lifecycle code. This PR does not notify listeners itself. It is the base for Wire #1271, #1272, #1273 and #1274; Queue #1725 maps `contextCount()` to Queue roll cycles and adds Queue context listeners.

## Review notes

`int` is intentional: the count is inherently bounded by Queue's `int` roll-cycle/context domain. All examined implementations and `ProgressiveContext.needsResending(int)` use `int`; no `long contextCount()` exists in this stack.

## Validation

- Java 8: `mvn -q clean -DskipTests javadoc:jar`
- Java 11: `mvn -q clean -DskipTests javadoc:jar`
- `mvn -q clean verify`
- Focused `DocumentContextLifecycleTest` - pass.
- All twelve reported Wire TeamCity configurations were successful at `e2077afab` during the detailed review.
- `git diff --check` - pass.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
@@ -243,6 +243,7 @@ public class BinaryWire extends AbstractWire implements Wire {
      */
     @Override
     public void reset() {
+        checkCanAdvanceOutputContext();
         writeContext.reset();
         readContext.reset();
         valueIn.resetState();
--- a/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/BinaryWire.java
@@ -248,6 +248,7 @@ public class BinaryWire extends AbstractWire implements Wire {
         valueIn.resetState();
         valueOut.resetState();
         bytes.clear();
+        advanceOutputContext();
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


tgd (approved):

src/main/java/net/openhft/chronicle/wire/BinaryWire.java
```diff
@@ -243,11 +243,13 @@ static boolean isDigit(char c) {
      */
     @Override
     public void reset() {
+        checkCanAdvanceOutputContext();
```
benbonavia, line 246: is this call required on every implementation `reset()`? it is already called in `advanceOutputContext()`

peter-lawrey:
Superseded by #1280, which contains the corrected final context-count and listener implementation in the single staged/develop delivery PR. This PR remains review provenance and must not be merged separately.

PR #1271, opened 2026-08-07
QUEUE-144: Add marshallable output context listener API

peter-lawrey:
## Summary

- Adds the shared `MarshallableOut.ContextListener` callback API.
- Adds `MarshallableOut.contextListener(...)` as an opt-in operation that defaults to unsupported.
- Adds the covariant `WireOut.contextListener(...)` entry point for fluent Wire configuration.

## Purpose and stack position

OpenHFT/Chronicle-Wire#1270 lets callers detect that an output context has changed, but it does not provide a way for an application to register context-writing behaviour. This PR adds that common registration contract so Queue, Wire and connection-based outputs can expose the same typed callback rather than defining product-specific listener interfaces.

The callback receives a preset method writer. This keeps context records on the normal serialization path and lets each output control its document and locking lifecycle. The default method remains unsupported so existing `MarshallableOut` implementations are source-compatible and can opt in explicitly.

This PR deliberately defines only the public API and its behavioural contract. It does not track contexts or invoke listeners. That separation keeps the API review focused before implementation is added by the remaining stack:

- OpenHFT/Chronicle-Wire#1270 provides the underlying `contextCount()` contract.
- OpenHFT/Chronicle-Wire#1272 implements listener notification lifecycle for Wire.
- OpenHFT/Chronicle-Wire#1273 enables the lifecycle in concrete `MarshallableOut` implementations.
- OpenHFT/Chronicle-Wire#1274 documents the completed Wire behaviour.
- OpenHFT/Chronicle-Queue#1725 implements Queue roll-cycle listeners using this shared API.

## Review focus

Review the callback timing, supplied-writer contract, failure and retry semantics, registration timing, and unsupported-by-default compatibility. Runtime lifecycle and output-specific behaviour belong to the later PRs.

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/MarshallableOut.java
+++ b/src/main/java/net/openhft/chronicle/wire/MarshallableOut.java
@@ -27,6 +27,22 @@ import java.util.stream.Stream;
 public interface MarshallableOut extends DocumentWritten, RollbackIfNotCompleteNotifier {
 
     /**
+     * Receives a callback when a {@link MarshallableOut} starts a new output context.
+     * The meaning of a context is implementation specific.
+     *
+     * @param <T> event interface type used by the supplied method writer
+     */
+    @FunctionalInterface
+    interface ContextListener<T> {
+        /**
+         * Called before the first application document is written in the new output context.
+         *
+         * @param writer preset method writer for writing context records
+         */
+        void onNewContext(T writer);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


tgd (approved):

benbonavia (approved):

peter-lawrey:
Superseded by #1280, which contains the corrected final context-listener API and implementation in the single staged/develop delivery PR. This PR remains review provenance and must not be merged separately.

PR #1272, opened 2026-08-07
QUEUE-144: Implement wire context listener lifecycle

peter-lawrey:
## Summary

- Implements the context-listener lifecycle for writable Wire implementations.
- Notifies the listener once per output context, immediately before the first data document and after any leading meta-data documents.
- Rearms notification on `Wire.reset()` while preserving the current context across `clear()`.
- Covers document-context and direct-write entry points, listener failure semantics, retained writers, and explicit `READ_ANY` rejection.

## Purpose and stack position

OpenHFT/Chronicle-Wire#1271 defines how an application registers a typed context listener, but deliberately does not decide when that listener runs. This PR supplies that lifecycle for Wire itself so the callback semantics can be reviewed and tested independently of Queue rolls, transport reconnects, and other output wrappers.

The lifecycle latches notification before invoking application code. A listener that writes partial context and then throws is therefore not retried in the same output context. `reset()` starts a new output context and rearms the listener; `clear()` only clears buffered data and does not resend context.

Listener-free wires use shared no-op states so the normal write path adds no per-document allocation. `READ_ANY` rejects listener registration because writes are delegated to an acquired underlying wire and a listener stored on the wrapper would otherwise be silently ignored.

This PR is the third change in the Wire stack:

- OpenHFT/Chronicle-Wire#1270 provides `contextCount()`, reset counting, and the progressive-context contract.
- OpenHFT/Chronicle-Wire#1271 adds the shared listener registration API.
- This PR implements listener notification for Wire.
- OpenHFT/Chronicle-Wire#1273 enables the lifecycle in concrete `MarshallableOut` implementations, including reconnect behaviour.
- OpenHFT/Chronicle-Wire#1274 documents the completed Wire contract.
- OpenHFT/Chronicle-Queue#1725 adds Queue roll-cycle semantics using the shared API.

## Review focus

Review the once-per-context timing, meta-data handling, reset versus clear behaviour, failure/no-retry contract, listener-free hot path, and coverage of write entry points. Output-wrapper and Queue-specific lifecycle decisions belong to the later PRs.

## Validation

- Java 8: `mvn -q clean -DskipTests javadoc:jar`
- Java 11: `mvn -q clean -DskipTests javadoc:jar`
- `mvn -q -Dtest=WireContextListenerLifecycleTest test`
- `mvn -q clean verify`
- `git diff --check`

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/AbstractAnyWire.java
+++ b/src/main/java/net/openhft/chronicle/wire/AbstractAnyWire.java
@@ -206,6 +206,15 @@ public abstract class AbstractAnyWire extends AbstractWire implements Wire {
         return wireAcquisition.acquireWire().acquireWritingDocument(metaData);
     }
 
+    // Writes are delegated to the acquired underlying wire, whose listener fields this setter would
+    // never reach - a listener registered here would be silently ignored, so fail loudly instead.
+    @NotNull
+    @Override
+    public <T> WireOut contextListener(@NotNull Class<T> writerType,
+                                       @NotNull MarshallableOut.ContextListener<? super T> listener) {
+        throw new UnsupportedOperationException("contextListener is not supported on a READ_ANY wire");
+    }
+
     @Override
     public String readingPeekYaml() {
         return wireAcquisition.acquireWire().readingPeekYaml();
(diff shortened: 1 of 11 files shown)


tgd (approved):

benbonavia (approved):

peter-lawrey:
Superseded by #1280, which contains the corrected fail-closed lifecycle and boundary implementation in the single staged/develop delivery PR. This PR remains review provenance and must not be merged separately.

PR #1273, opened 2026-08-07
QUEUE-144: Support context listeners in marshallable outputs

peter-lawrey:
## Summary

- Enables `contextListener(...)` for append-mode file, string-consumer, and HTTP `MarshallableOut` implementations.
- Treats append files and string consumers as one logical output context across buffer clears.
- Treats each top-level HTTP document as a new output context, so every independent POST contains the context needed to interpret it.
- Rejects context listeners for overwrite-mode files because each document replaces the file and would discard context written by an earlier document.
- Tests complete serialised output with representative context and event DTOs.

## Purpose

OpenHFT/Chronicle-Wire#1272 implements once-per-output-context listener notification on writable Wire instances. The wrappers changed here must define what an output context means for their destination and pass listener registration to the backing Wire.

Without this adapter-specific handling, HTTP can send later standalone requests without the context required to interpret them, while overwrite-mode files can appear to support context listeners even though the surviving file no longer contains that context.

## Lifecycle decisions

- **String consumer:** buffer clearing does not create a new output context, so context is emitted once before the first event.
- **Append file:** appended documents share one output context and retain the first context record in the file.
- **Overwrite file:** context listeners are rejected with an explanatory exception.
- **HTTP:** each top-level writable document represents one POST. Its one-based context count advances only when that document opens, remains unchanged when it closes, and does not advance for a nested acquired document. After a successful POST, the backing Wire is reset so its listener is rearmed before the next document.

## Stack position

This PR is stacked on OpenHFT/Chronicle-Wire#1272 and deliberately contains only concrete `MarshallableOut` adapter behaviour. It does not change the shared listener API or the core Wire lifecycle.

OpenHFT/Chronicle-Wire#1274 documents the completed Wire contract, and OpenHFT/Chronicle-Queue#1725 applies the same API to Queue roll cycles.

## Review focus

Review the chosen output-context boundaries, HTTP open/close counting, reset-versus-clear behaviour, overwrite-file rejection, and the full-output assertions.

## Validation

- `mvn -q -Dtest=DocumentContextLifecycleTest,MarshallableOutContextListenerTest test`
- `mvn -q clean verify`

### DIFF
--- a/src/main/java/net/openhft/chronicle/wire/internal/FileMarshallableOut.java
+++ b/src/main/java/net/openhft/chronicle/wire/internal/FileMarshallableOut.java
@@ -116,4 +116,17 @@ public class FileMarshallableOut implements MarshallableOut {
     static class FMOOptions extends SelfDescribingMarshallable {
         boolean append; // Indicates if data should be appended to the existing file
     }
+    @Override
+    public <T> MarshallableOut contextListener(Class<T> writerType,
+                                               MarshallableOut.ContextListener<? super T> listener) {
+        // In overwrite mode every document atomically replaces the file, so context records written
+        // with the first document would be lost from the surviving output - fail loudly instead.
+        if (!options.append)
+            throw new UnsupportedOperationException(
+                    "contextListener requires append mode (add ?append=true to the file URL): "
+                            + "in overwrite mode each document replaces the file, discarding the context records");
+        wire.contextListener(writerType, listener);
+        return this;
+    }
+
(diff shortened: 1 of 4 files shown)


tgd (approved):

benbonavia (approved):

peter-lawrey:
Closed as deferred rather than merged. #1280 retains the supported Wire listener boundary while excluding the deferred adapter scope from this release.

PR #1274, opened 2026-08-07
QUEUE-144: Document wire context listener contract

peter-lawrey:
Documents the Wire context-listener and context-count contract in the decision log and interface-control documentation.

## Why this contract exists

Some streams carry a large, stable description followed by many small value updates. Repeating the description in every document is expensive; emitting it only once globally makes independently retained or delivered contexts impossible to interpret. The contract allows an application to write context once per output-defined context - such as a Wire generation, Queue roll, HTTP request, or connection generation - so compact data remains interpretable at those boundaries.

## Direction, not scope

A future telemetry layer could use this primitive to emit source or series definitions and projection checkpoints at output-context boundaries, then write compact value events that refer to them. This PR does not define that layer, its identifiers, or its replay and checkpoint policy. It documents only the generic lifecycle needed by Queue, FIX, Services, and similar consumers.

## Review Notes

Stacked on #1273. This is kept last so implementation reviewers can focus on code before reviewing contract prose.

### DIFF
--- a/src/main/docs/decision-log.adoc
+++ b/src/main/docs/decision-log.adoc
@@ -55,3 +55,41 @@ Impact & Consequences ::
 Notes/Links ::
 ** `operational-requirements.adoc`
 ** `functional-requirements.adoc`
+
+=== [FN-008] MarshallableOut Context Listener
+
+Date :: 2026-07-23
+Context ::
+* Chronicle Queue needed to write a per-roll-file preamble before the first application excerpt,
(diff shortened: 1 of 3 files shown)


tgd (approved):

benbonavia (approved):

peter-lawrey:
Superseded by #1280, which carries the corrected compact context-listener documentation with the final implementation. This PR remains review provenance and must not be merged separately.

PR #1275, opened 2026-08-20
QUEUE-143: Scope EOF replacement to validated recovery

peter-lawrey:
## Why

Exact-index replication can need to restore authoritative data into an older cycle that the target sealed from incomplete local state. Normal writes must not acquire the same permission: an EOF marker can also represent an archived, deliberately sealed, or corrupt store.

## What changed

- Ordinary `enterHeader` calls remain strict and continue to throw `WriteAfterEOFException` at EOF.
- `InternalWire.recoverFromEndOfData()` is the explicit recovery-only primitive. It reports the replaced byte position and is not exposed through generic `WireOut`.
- The decision log requires Queue to hold its write lock and validate the exact recovery index before invoking the primitive.
- Wire no longer claims that EOF is an unconditional soft marker or that `WriteAfterEOFException` is legacy.
- Focused tests distinguish an ordinary rejected write from explicit recovery and verify later documents remain readable.

## Coordination and compatibility

This change is intended to ship only with OpenHFT/Chronicle-Queue#1732 and ChronicleEnterprise/Chronicle-Queue-Enterprise#945. Queue owns exact-index validation, contextual warnings, dirty-cycle tracking, and EOF restoration; CQE owns the replication trigger and peer context. The supported BOM must be updated only after that combination passes.

Older Queue versions never call the new internal operation and therefore remain strict with this Wire version.

## Validation

- `mvn -q -o -Dtest=WriteOverEOFTest test`
- Queue recovery, normalisation, and ordinary-write focused tests
- CQE production-path A/B/C replication recovery test
- `git diff --check` across all three repositories

Depends on the coordinated Queue and CQE changes above.

### DIFF
--- a/src/main/docs/decision-log.adoc
+++ b/src/main/docs/decision-log.adoc
@@ -103,8 +103,15 @@ Impact & Consequences ::
   Chronicle Queue catches it for both ordinary document headers and sequential
   byte appends, rolls forward without replacing EOF, and does not expose it
   through its ordinary appender APIs.
-* The internal recovery operation reports whether and where it replaced the
-  marker; the coordinating storage layer emits the contextual warning.
+* The internal recovery operation performs one compare-and-swap at the current
+  write position and reports whether it replaced the marker. It does not search
+  or reposition. The coordinating storage layer validates and positions the
+  Wire first, then relies on the CAS result without a preliminary header read.
+  EOF is used with padded Wires, so the caller aligns the selected position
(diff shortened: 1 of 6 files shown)


peter-lawrey:
Superseded by #1280. The integration PR retains only the minimal EOF header-state cleanup and hard-seal contract; the abandoned Wire recovery primitive is not included.

PR #1280, opened 2026-08-27
QUEUE-144: integrate context listeners with QUEUE-146 EOF boundary

peter-lawrey:
## Purpose

Provide one `staged/develop` integration PR to `develop` for the Wire behaviour required by QUEUE-144 plus the small QUEUE-146 EOF-state correction.

The branch contains the accepted context-listener implementation and the focused EOF boundary. It excludes the abandoned Wire recovery primitive; Queue owns rolling and exact-index recovery.

## QUEUE-144 context listeners

- one notification boundary for generated, reflective and explicit-proxy method writers;
- the supplied writer is bound to a frozen internal output;
- incomplete, rolled-back or structurally uncommitted listener output fails closed;
- direct-write failures roll back and poison the lifecycle until reset;
- Binary, Text, YAML and RawWire lifecycle parity;
- harmless `DocumentContext.NOOP`;
- one canonical `MarshallableOut.UNSET_CONTEXT` value for unavailable context counts, shared by Wire, document-context wrappers, `READ_ANY` and downstream outputs while zero remains valid.

Interface-level `@Comment` emission is deliberately not added to the Text/YAML proxy path: no supported interface uses that combination, and it is unrelated to EOF handling. A short `//!` note records the scope decision.

## QUEUE-146 EOF boundary

- `enterHeader()` clears only transient `insideHeader` state before propagating `WriteAfterEOFException`;
- the observed EOF marker and write position are unchanged;
- a repeated attempt reports the same seal rather than a stale nested-header error;
- Wire exposes no EOF replacement operation.

## Exact head and validation

Current head: `57fb8ee628cdc1a4ce86a012f635a787ad28cf55`.

Validated code head: `3d72ec2c3f537f5888e16d6c44f0f95fde077b85`. The current-head delta only documents the 144-character wrapping convention in `AGENTS.md`.

- Focused `DocumentContextLifecycleTest` and `WireContextListenerLifecycleTest` run passed.
- Full `mvn verify`: 11,398 tests, 0 failures/errors, 298 skipped.
- Checkstyle: 0 violations; Javadocs and binary compatibility passed.
- Review notes cover the behavioural release hunks; the additive constant's note explains why a public symbol is needed even though its runtime value is unchanged.

This PR supersedes #1281, #1275 and #1282 as delivery vehicles; those remain review provenance only. Immutable release/BOM coordinates, exact-head CI and fresh review remain gates.


src/main/java/net/openhft/chronicle/wire/ActiveWireContextListenerLifecycle.java
```diff
@@ -0,0 +1,216 @@
+/*
+ * Copyright 2013-2025 chronicle.software; SPDX-License-Identifier: Apache-2.0
```
tgd, line 2: Change copyright upper bound to 2026 if possible

src/main/java/net/openhft/chronicle/wire/ActiveWireContextListenerLifecycle.java
```diff
@@ -0,0 +1,216 @@
...
+import net.openhft.chronicle.core.Jvm;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Stateful lifecycle allocated only for a configured context listener.
+ *
+ * <p>A failed callback poisons the current context. It is not retried because it may already have
+ * committed context records, and application data is rejected until the wire is reset.</p>
+ */
+final class ActiveWireContextListenerLifecycle implements WireContextListenerLifecycle {
```
tgd: Should we make statement about thread safety of this? Should we enforce single threaded access?

src/main/java/net/openhft/chronicle/wire/ActiveWireContextListenerLifecycle.java
```diff
@@ -0,0 +1,216 @@
...
+        Jvm.warn().on(ActiveWireContextListenerLifecycle.class,
+                "Context listener failed: writerType=" + writerType.getName()
+                        + ", listenerType=" + listener.getClass().getName()
+                        + ", outputType=" + wire.getClass().getName()
+                        + ", outputIdentity=0x" + Integer.toHexString(System.identityHashCode(wire))
+                        + ", contextCount=" + wire.contextCount(),
+                throwable);
+    }
+
+    @SuppressWarnings({"rawtypes", "unchecked"})
```
tgd, line 158: Is it worth adding a type parameter to `ActiveWireContextListenerLifecycle` to match the type of the embedded listener to avoid this cast

src/main/java/net/openhft/chronicle/wire/MarshallableOut.java
```diff
@@ -26,6 +26,24 @@
...
+    /**
+     * Receives a callback when a {@link MarshallableOut} starts a new output context.
+     * The meaning of a context is implementation specific.
+     *
+     * @param <T> event interface type used by the supplied method writer
+     */
+    //! WireContextListenerLifecycleTest#contextListenerWaitsForDataAfterMetadataAndWritesDtoOnce
+    //! demonstrates the callback contract and preset method writer exposed by this API.
+    @FunctionalInterface
+    interface ContextListener<T> {
```
tgd, line 47: Consider tagging this with an experimental or preview annotation in case this interface evolves during this release cycle - gives us a little more freedom around deprecation cycle by notifying folks to not depend on it until stabilised

src/main/java/net/openhft/chronicle/wire/MarshallableOut.java
```diff
@@ -36,6 +54,32 @@ static MarshallableOutBuilder builder(URL url) {
...
+     * count while {@link Wire#clear()} leaves it unchanged. Queue implementations return the roll
+     * cycle, and connection-based outputs should return a one-based connection count.
+     * Implementations may throw {@link IllegalStateException} when the output context is not yet
+     * known, such as when using double buffering. Implementations whose count cannot advance without
+     * overflowing must throw {@link IllegalStateException} before opening the next output context
+     * rather than wrap into the negative range.
+     *
+     * @return the output context count, or a negative value if unavailable
+     */
+    default int contextCount() {
```
tgd, line 86: The semantics of contextCount don't ring quite right to me yet - -1 indicates does not expose an output, writeable wire starts at 0, but the name contextCount indicates this is a *count* not a *code* - I am not sure whether its a count or a code

src/main/java/net/openhft/chronicle/wire/WireInternal.java
```diff
@@ -138,37 +138,54 @@ public static <E extends Enum<E>> E internEnum(@NotNull Class<E> eClass, @NotNul
      */
     public static long writeData(@NotNull WireOut wireOut, boolean metaData, boolean notComplete,
                                  @NotNull WriteMarshallable writer) throws InvalidMarshallableException {
+        //! WireContextListenerLifecycleTest#wiresWriteDataUsesLifecycleAndClosesRegistrationWindow
+        //! and #directWriteEntryPointsInvokeListenerOnce demonstrate that both low-level helpers and
+        //! public complete/not-complete direct writes share one metadata-aware notification boundary.
+        if (wireOut instanceof AbstractWire)
```
tgd, line 144: There are a few instances of this instance of check, is it possible to handle this with a different way of modelling things to avoid the instanceof

src/main/java/net/openhft/chronicle/wire/ActiveWireContextListenerLifecycle.java
peter-lawrey: Wire it's user is single threaded, as this class is package local so not used directly.

src/main/java/net/openhft/chronicle/wire/ActiveWireContextListenerLifecycle.java
peter-lawrey: That's a big one. All the files need it really, didn't want to complicate the PRs. There is abotu 10 in the chain so far.

src/main/java/net/openhft/chronicle/wire/MarshallableOut.java
peter-lawrey: It is a zero indexed count. First is 0.

src/main/java/net/openhft/chronicle/wire/WireInternal.java
peter-lawrey: The problem is these are internal APIs. Using an interface would help the instanceof, but should make it public.

src/main/java/net/openhft/chronicle/wire/MarshallableOut.java
peter-lawrey: It is zero indexed because the main (perhaps only real) use case is Chronicle Queue which starts at epoch on roll cycle 0

tgd (approved):

PR #1284, opened 2026-09-03
Re-enable fast Wire regression tests

peter-lawrey:
Re-enable fast Wire regression tests and repair their fixtures and assertions. JSON/YAML round trips, Unicode values, RawWire byte consumption, String maps, DTO conversion and document-context handling gain checks of decoded values, callback order and resource ownership.

**Depends on #1286.** The Boolean-looking String writer/reader correction, String-key fixture and both Boolean regression classes belong to that prerequisite. This PR targets `fix/wire-quoted-boolean-strings`, so they are excluded from its own diff. Merge #1286 first, then retarget this PR to `develop`.

The remaining release-code changes are:

- Quote dot-prefixed numeric-looking Strings in `YamlWire`, demonstrated by `ScalarTypePreservationTest#dotPrefixedNumericTextRetainsType`.
- Distinguish nested YAML document acquisition from continuation of an explicitly chained document, demonstrated by `WriteDocumentContextTest#nestedYaml` and `#chainedYaml`.

The test changes restore disabled format/lifecycle scenarios, assert decoded Unicode fields, verify RawWire callback values/order, check DTO element types, tighten expected-negative JSON parsing, repair compatibility-fixture selection/comparisons, and close resources used by these paths. The numeric-scalar test remains separate from #1286's Boolean regressions.

Validation on 14 September 2026, exact head `23945e6e3c0ba1026ae6fd5605f48dc1dfe003bb`, above #1286 at `70fbf328cb6f18259814d21647acb96d4f4a333b`:

- The complete `mvn -B -ntp -Dsurefire.rerunFailingTestsCount=0 clean verify` passes on OpenJDK 21.0.12: **11,429 reported tests, 0 failures, 0 errors, 62 existing skips and no retries**.
- All 23 inherited Boolean regression cases execute without skips, including the 15 newly parameterised declared-container cases. The numeric-scalar and YAML document-context regressions also execute.
- The restack preserves this PR's complete own binary patch byte-for-byte: **13 files, +245/-170**. Existing contributor history is retained; no behaviour from the prerequisite is delivered a second time.
- Packaging, Javadocs, licence checks and binary compatibility verification pass. Maven 3.9.11, Linux amd64, an isolated offline dependency cache and four CPU cores were used. The initial combined run passed all tests but failed the parent's new-file licence-header check; that header was corrected before this final clean run.

The broader slow, performance and known thread-unsafe scenarios remain outside this work. Required current-head CI and review remain merge gates.


src/main/java/net/openhft/chronicle/wire/YamlWireOut.java
```diff
@@ -347,6 +348,11 @@ protected Quotes needsQuotes(@NotNull CharSequence cs) {
         if (cs.length() == 0)
             return Quotes.DOUBLE;
 
+        //! ScalarTypePreservationTest#textAndTypedScalarsRetainTypes demonstrates that bare "true" and "false" Strings are read as
+        //! Booleans. Quoting every String would alter established text output, so quote only these resolver-sensitive spellings.
+        if (StringUtils.isEqual(cs, "true") || StringUtils.isEqual(cs, "false"))
```
tgd, line 353: There is a functionality change in here we should call out with a Jira / release note

tgd (approved):

src/main/java/net/openhft/chronicle/wire/YamlWireOut.java
peter-lawrey: This has been extracted to it's own PR #1286

benbonavia (approved):

PR #1286, opened 2026-09-09
CORE-79: Preserve boolean-looking Strings in Text, YAML and JSON

peter-lawrey:
String values `"true"` and `"false"` could come back as Booleans after textual Wire round trips. Quote those exact String spellings when writing Text/YAML and honour the original quotes when inferring scalar types in Text/YAML/JSON readers. Typed Boolean values keep their Boolean representation.

This extracts the behaviour identified in [the review of #1284](https://github.com/OpenHFT/Chronicle-Wire/pull/1284#discussion_r3962098450). The writer correction needs both reader guards: quoting alone does not stop the old readers from inferring a Boolean. String field names/map keys with these spellings are also quoted. Consumers comparing encoded output or relying on the old quoted-string inference must account for this compatibility change.

The added `release-notes.adoc` is removed following [the release-process review](https://github.com/OpenHFT/Chronicle-Wire/pull/1286#discussion_r4001621760). The compatibility explanation is retained here.

The [nested-type review](https://github.com/OpenHFT/Chronicle-Wire/pull/1286#discussion_r3989482977) has two complementary layers of coverage:

- `BooleanScalarTypePreservationTest` retains all eight existing tests: scalar inference, mixed lists, nested maps/lists and siblings, independently supplied quoted flow/block input, Boolean-looking String keys and fresh nested DTO `Object` fields.
- `BooleanContainerFieldTypePreservationTest` adds five tests, each parameterised separately for `TEXT`, `YAML_ONLY` and `JSON`: **15 additional cases**. Declared DTO fields exercise the array, collection and map field-reading paths in `WireMarshaller`; simply adding deeper untyped nesting would not cover that boundary.

| New test | Distinct evidence |
| --- | --- |
| `setFieldDoesNotCollapseDistinctScalarTypes` | `Set<Object>` retains all four distinct String/Boolean values; the old code collapses them to two. |
| `handWrittenSetInputPreservesAllFourValues` | The same four-element contract using independent input, without invoking a Wire writer. |
| `objectArrayFieldRetainsScalarTypes` | `Object[]` element values/types plus the String and Boolean sibling fields following the array. |
| `mapOfListsFieldRetainsScalarTypes` | A declared `Map<String, List<Object>>` with two oppositely ordered mixed lists. |
| `listOfMapsFieldRetainsScalarTypes` | A declared `List<Map<String, Object>>` containing two maps with different scalar ordering. |

The decoded DTO fields initially contain null. Collection assertions compare values and types through equality without requiring a specific implementation. Hand-written Text input retains separator whitespace so a syntax limitation cannot substitute for the intended scalar/cardinality failure.

Validation on 14 September 2026, Linux amd64, exact head `70fbf328cb6f18259814d21647acb96d4f4a333b`:

| Clean verification | Reported tests | Existing skips | Failures/errors/retries |
| --- | ---: | ---: | ---: |
| Java 1.8.0_502 | 11,426 | 294 | 0 |
| Java 21.0.12 | 11,426 | 298 | 0 |

All **23 focused cases** execute and pass without skips on both JDKs. Copying the identical final test classes onto pre-fix source `42879cf5979950b5176efe368944eab1e6aa6c13`, with identical dependency JAR hashes on Java 21, produces **23 assertion failures, no errors/skips/retries**. All 15 new format-specific cases independently detect the defect: set cardinality, array element types or mixed-container equality. The existing eight tests retain their earlier first-failing-subcase limitation.

Commands: `mvn -B -ntp -Dsurefire.rerunFailingTestsCount=0 clean verify`, or add `-Dtest=BooleanScalarTypePreservationTest,BooleanContainerFieldTypePreservationTest` and use `clean test` for the focused selection. Local runs used Maven 3.9.11, an isolated offline dependency cache and four CPU cores per process. Packaging, Javadocs, licence checks and binary compatibility verification pass. The initial full attempts passed their tests but failed the new file's licence-header check; that header was corrected before these final clean runs.

Current `develop` at `4a6acae49271e3fc738505fe12d7ce6a05fe7dd0` is already an ancestor. [#1284](https://github.com/OpenHFT/Chronicle-Wire/pull/1284) remains the dependent PR and is restacked with its own binary patch unchanged. Its complete Java 21 verification passes at `23945e6e3c0ba1026ae6fd5605f48dc1dfe003bb`: 11,429 reported tests, 62 existing skips, no failures/errors/retries. Integrate this prerequisite first. Required current-head CI and review remain merge gates.


src/test/java/net/openhft/chronicle/wire/BooleanScalarTypePreservationTest.java
```diff
@@ -0,0 +1,65 @@
...
+import net.openhft.chronicle.bytes.Bytes;
+import org.jetbrains.annotations.NotNull;
+import org.junit.Test;
+
+import java.util.function.Consumer;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertSame;
+
+public class BooleanScalarTypePreservationTest extends WireTestCommon {
```
tgd, line 15: Suggest adding test coverage for nested types as well

tgd (approved):

src/main/docs/release-notes.adoc
```diff
@@ -0,0 +1,37 @@
+= Chronicle Wire Release Notes
```
benbonavia: this file is not required - we have a release notes process

benbonavia (changes requested):
Nit picking - please remove the release-notes.adoc file

benbonavia (approved):

